PR #105, opened 2024-10-28
lite local setup

dinmukhamedm:
- Introduce a runtime control variable `FULL_BUILD` that disables some reliance on other services:
   - App server:
       - Clickhouse is never connected, and query time we return empty data
       - RabbitMQ is replaced with synchronously writing span and trace to DB from producer (ingestor) side
       - SemanticSearchService is replaced with dummy client returning default responses
   - Frontend
       - Remove dashboard (clickhouse-dependent) from navbar
- `FULL_BUILD` defaults to true unless explicitly set to `false`, so that this does not affect Prod
- Add docker compose for lite build and docker-compose-full for full build
- Changes in README and CONTRIBUTING to reflect the above
- Remove reliance on `LanguageModelRunner` from LLM call cost computation. This slightly reduces freedom for custom logic like we used to have in Gemini, but makes code a lot cleaner.
- Migrations are now handled by drizzle, so no custom image for postgres is build. This is controlled by `LOCAL_DB` feature, which is enabled always if `ENVIRONMENT=production` is not set.
- edit image route from `/api/images/[projectId]/[imageId]` to `/api/projects/[projectId]/images/[imageId]` to be consistent with other routes structure. The old route will redirect to the new one for now, because many spans still have the older route as the URL of the image


skull8888888 (approved):
tiny pr, had a quick look

PR #208, opened 2024-11-14
fix page number updated to 0 when changing filter and updated custom …

### LINKED ISSUE #150: Reset pageNumber to 0 when changing filters
This is true for any table with [manual] pagination, spans and traces are a good candidate

### Reproduction step
- Have multiple pages of traces for the last 24 hours
- Have less than one page of traces for the last 1 hour
- Set the filter to 24 hours.
- Go past page 1, e.g. page 3
- Set the filter back to 1 hour

### Expected behavior
I can see my traces for the last hour

### Actual behavior
I see an empty page, because I am in page 3/1

TODO:
Reset pageNumber to 0 when filters change. We can be smart here, e.g. do this only if filters change or if the time range _decreases_. If we merely increase the time range, it probably makes sense to stay on the same page

sarthakvijayvargiya:
fix #150

-  Fixed --  page number update to 0  when changing filter
- Fixed -- page number update to 0 when adding custom date filter

https://github.com/user-attachments/assets/8417d67e-f3bb-4d7e-969f-179c2c1083e1

----

> [!IMPORTANT]
> Reset `pageNumber` to `0` when changing filters in `date-range-filter.tsx`.
> 
>   - **Behavior**:
>     - Reset `pageNumber` to `0` when changing filters in `AbsoluteDateRangeFilter` and `DateRangeFilter`.
>     - Specifically, reset `pageNumber` when setting `startDate`, `endDate`, or `pastHours`.
>   - **Files**:
>     - Modifications in `date-range-filter.tsx` to handle `pageNumber` reset logic.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=lmnr-ai%2Flmnr&utm_source=github&utm_medium=referral)<sup> for ace20664382df333da27c6206224b9dffbd3b141. It will automatically update as commits are pushed.</sup>


frontend/components/ui/date-range-filter.tsx
```diff
@@ -177,6 +177,8 @@ function AbsoluteDateRangeFilter() {
               }
               onClick={() => {
                 searchParams.delete('pastHours');
+                searchParams.set('pageNumber', '0');
+                searchParams.set('pageNumber','0');
```
dinmukhamedm: nit: this is duplicated, remove line 181

frontend/pnpm-lock.yaml
```diff
@@ -11255,4 +11255,4 @@ packages:
 
   /zwitch@2.0.4:
     resolution: {integrity: sha512-bXE4cR/kVZhKZX/RjPEflHaKVhUVl85noU3v6b8apfQEc1x4A+zBxjZ4lN8LqGd6WZ3dl98pY4o717VFmoPp+A==}
-    dev: false
+    dev: false
```
dinmukhamedm, line 11258: nit: this is redundant, remove the diff. Keep the trailing newline

frontend/components/ui/date-range-filter.tsx
```diff
@@ -230,6 +231,8 @@ export default function DateRangeFilter() {
                   searchParams.delete('endDate');
                   searchParams.delete('groupByInterval');
                   searchParams.set('pastHours', range.value);
+                  searchParams.delete('pageNumber');
+                  searchParams.set('pageNumber', '0');
```
dinmukhamedm: ditto ellipsis comment, `delete` is redundant

dinmukhamedm (changes requested):
Left a few nit comments

Issue #140, opened 2024-11-02
Migrate pipelines queries to Drizzle


dinmukhamedm:

alviskalev:
hi @dinmukhamedm could you please assign this issue to me?

PR #239, opened 2024-11-23
refac: Migrate all pipelines db operation to drizzle on NextJs

alviskalev:
closed https://github.com/lmnr-ai/lmnr/issues/140

### DIFF
--- a/frontend/app/api/projects/[projectId]/pipelines/[pipelineId]/route.ts
+++ b/frontend/app/api/projects/[projectId]/pipelines/[pipelineId]/route.ts
@@ -1,25 +1,24 @@
-import { authOptions } from '@/lib/auth';
-import { fetcher } from '@/lib/utils';
-import { getServerSession } from 'next-auth';
+import { db } from '@/lib/db/drizzle';
+import { and, eq } from 'drizzle-orm';
+import { pipelines } from '@/lib/db/migrations/schema';
+import { z } from 'zod';
 
 export async function DELETE(
   req: Request,
   { params }: { params: { projectId: string; pipelineId: string } }
 ): Promise<Response> {
-  const projectId = params.projectId;
   const pipelineId = params.pipelineId;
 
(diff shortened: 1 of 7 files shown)


frontend/app/api/projects/[projectId]/pipelines/[pipelineId]/route.ts
```diff
@@ -29,37 +28,62 @@ export async function GET(
...
+      status: 404
+    });
+  }
+
+  return new Response(JSON.stringify(pipelineRes), { status: 200 });
 }
 
+const UpdatePipelineSchema = z.object({
+  name: z.string().optional(),
+  visibility: z.enum(['PUBLIC', 'PRIVATE']).optional()
```
Comment (dinmukhamedm) [inline frontend/app/api/projects/[projectId]/pipelines/[pipelineId]/route.ts:46]: I think there is no way to update the "visibility" part from the UI. Let's remove handling it, this is sort of legacy. Sorry, I should have mentioned that in the issue description.

frontend/lib/pipeline/types.ts
```diff
@@ -279,3 +290,6 @@ export const PROVIDERS = providerMapping.map((provider) => provider.provider);
 export const LANGUAGE_MODELS = providerMapping.flatMap(
   (provider) => provider.models
 );
+export const DEFAULT_NEW_PIPELINE_VERSION_ID_STRING =
+  'db6d1708-9836-42f2-a3ea-732ca7709039';
```
dinmukhamedm, line 294: nit, let's rename this to `DEFAULT_NEW_PIPELINE_TEMPLATE_ID`.
Two small changes:
- clearer communication that this is a template. Yes, I gave the variable the original name and only now I realize how confusing it is.
- no need to call it a string. The only reason it was called a string was an indication in Rust to parse it to UUID. As we don't need to do that in JS, we can drop that part

dinmukhamedm (commented):
As a draft this looks good to me overall. Let's go ahead and test this/remove the redundant routes and methods in the backend

PR #270, opened 2024-12-06
Refactor datatable component to include filter clearing functionality

### LINKED ISSUE #269: Add CTA to reset filters if filters are applied and 0 results are shown
Any page with `DatatableFilter` is sometimes misleadingly empty when filters are applied – i.e. it is easy to accidentally forget that filters are applied.

Add some kind of a `Card` right in the middle of the page with a call to action to disable filters to see all results. The text could be something like

"Applied filters returned no results. <Button> Clear filters </Button>".

This will probably involve editing either the `DataTable` component or pages that use it

devansh-m12:
fixes #269 

@dinmukhamedm I've updated the DataTable component to check if a filter exists in the search parameters and is applied. Only then do we have the permission to clear filters."

Screenshot 2024-12-06 at 8 48 10 AM

----

> [!IMPORTANT]
> Adds filter clearing functionality to `DataTable` component, allowing users to clear applied filters via a button.
> 
>   - **Behavior**:
>     - Adds `clearFilters` function in `datatable.tsx` to remove 'filter' from search parameters and update the URL.
>     - Displays 'Clear filters' button in `TableCell` when filters are applied and no results are found.
>   - **Imports**:
>     - Adds `usePathname`, `useRouter`, and `useSearchParams` from `next/navigation` in `datatable.tsx`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=lmnr-ai%2Flmnr&utm_source=github&utm_medium=referral)<sup> for 8f65322987077538fb5507643e1c9c82f19ef741. It will automatically update as commits are pushed.</sup>


frontend/components/ui/datatable.tsx
```diff
@@ -284,7 +298,12 @@ export function DataTable<TData>({
                 colSpan={columns.length}
                 className="text-center p-4 text-secondary-foreground"
               >
-                No results
+                {searchParams.get('filter') !== null ? 'Applied filters returned no results. ' : 'No results'}
+                {searchParams.get('filter') !== null && (
+                  <Button className="text-primary" variant={'ghost'} onClick={clearFilters}>
+                    Clear filters
+                  </Button>
```
dinmukhamedm: This leaves a big visual gap to the left of this text. Let's make this a `<span>` with a `"hover:cursor-pointer"` className.

dinmukhamedm (changes requested):

frontend/components/ui/datatable.tsx
```diff
@@ -92,6 +92,20 @@ export function DataTable<TData>({
...
   const [expandedRows, setExpandedRows] = useState<ExpandedState>({});
 
+  const searchParams = new URLSearchParams(useSearchParams().toString());
+  const pathName = usePathname();
+  const router = useRouter();
+
+  const clearFilters = () => {
+    if (searchParams.size > 0) {
+      // clear all filters
+      if (searchParams.get('filter') !== null) {
```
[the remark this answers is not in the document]
dinmukhamedm: This is intentional, we don't clear the other params

dinmukhamedm: I'm not sure the strict check for null is correct, can we just do ` != null ` so that it matches undefined as well? Just in case

dinmukhamedm:
@devansh-m12 LGTM. Can you please attach a screenshot of what it looks like after your last change?

devansh-m12:
Here, @dinmukhamedm I've implemented the requested changes. To clarify, it removes the filter from the search parameters but does not reset the filter fields, allowing previously applied filters to be re-applied with a single click. you can check recording for refrence.

Screenshot 2024-12-06 at 9 31 11 AM



https://github.com/user-attachments/assets/3b4c92c4-3e27-4928-bb0f-0df269df946a

dinmukhamedm:
Awesome, that works great, merging to dev now!

dinmukhamedm (approved):

PR #273, opened 2024-12-06
frontend: expandable sidebar

adilb99:
video of how it looks:
https://vimeo.com/1036801461/fe4c0a66c2?share=copy

A very hacky solution to having an expandable sidebar, but it works well on local so far. To do:

- [x] maybe a more robust check for whether it's a project page or a root page @ frontend/components/ui/header.tsx:26-28 (or use a different header component in the project page to have the SidebarTrigger properly in the header without affecting the root layout page).
- [x] less hacky styling cause currently the sidebar.tsx uses some of the styles from the old navbar directly instead of a reusable way.
- [x] resolve conflicts?
- [ ] apply ESLint rules consistently

----

> [!IMPORTANT]
> Introduces an expandable sidebar with mobile support, using new `Sidebar` components and updated styling across the frontend.
> 
>   - **Behavior**:
>     - Introduces expandable sidebar using `SidebarProvider` and `Sidebar` components in `layout.tsx` and `project/[projectId]/layout.tsx`.
>     - Sidebar can be toggled using `SidebarTrigger` in `header.tsx`.
>     - Mobile support added with `useIsMobile` hook in `use-mobile.tsx`.
>   - **Components**:
>     - New `Sidebar` components in `sidebar.tsx` for layout and interaction.
>     - `ProjectNavbar` updated to use `Sidebar` components in `project-navbar.tsx`.
>     - `Sheet` components in `sheet.tsx` used for mobile sidebar.
>   - **Styling**:
>     - New CSS variables for sidebar styling in `globals.css`.
>     - Tailwind configuration updated in `tailwind.config.js` and `tailwind.config.ts` to support new styles.
>   - **Misc**:
>     - Minor refactoring in `pipeline-version-context.tsx` and `project-context.tsx` for consistency.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=lmnr-ai%2Flmnr&utm_source=github&utm_medium=referral)<sup> for 59c642afe47e917022f0e36c7b0e646d286c9339. It will automatically update as commits are pushed.</sup>


frontend/app/globals.css
```diff
@@ -70,6 +70,14 @@
...
     --chart-4: 280 65% 60%;
     --chart-5: 340 75% 55%;
+    --sidebar-background: 0 0% 98%;
+    --sidebar-foreground: 240 5.3% 26.1%;
+    --sidebar-primary: 240 5.9% 10%;
+    --sidebar-primary-foreground: 0 0% 98%;
+    --sidebar-accent: 240 4.8% 95.9%;
+    --sidebar-accent-foreground: 240 5.9% 10%;
+    --sidebar-border: 220 13% 91%;
+    --sidebar-ring: 217.2 91.2% 59.8%;
```
dinmukhamedm, line 80: are all these additional colors required by shadcn's `<Sidebar>`. Otherwise, I don't think we need them all, and could just tailwind all the necessary colors on the sidebar

frontend/assets/logo/laminar.svg
dinmukhamedm, line 1: we'll add the new logo anw I guess

frontend/components/project/project-navbar.tsx
```diff
@@ -37,52 +56,75 @@ export default function ProjectNavbar({ projectId }: ProjectNavBarProps) {
...
-          </Link>
-        ))}
-      </div>
-      <div className="flex-grow"></div>
-      <div className="mb-8">
-        <AvatarMenu />
-      </div>
-    </div>
+    <Sidebar className="border-r" collapsible='icon'>
+      <SidebarHeader className="h-14 bg-background">
```
dinmukhamedm, line 93: I'd say we center this: `mx-auto`? But again, let's wait for Robert's input

frontend/components/project/project-navbar.tsx
```diff
@@ -37,52 +56,75 @@ export default function ProjectNavbar({ projectId }: ProjectNavBarProps) {
...
+                  <option.icon className="flex justify-center items-center !w-[20px] !h-[20px]" />
+                  <span>{option.name}</span>
+                </Link>
+              </SidebarMenuButton>
+            </SidebarMenuItem>
+          ))}
+        </SidebarMenu>
+        <div className="flex-grow" />
+        <div className="p-2">
+          <AvatarMenu />
```
dinmukhamedm: avatar in the corner of an expanded sidebar looks a bit too empty. Should we maybe add an email to the right of it?

frontend/components/ui/sidebar.tsx
```diff
@@ -0,0 +1,745 @@
...
+          (value: boolean | ((value: boolean) => boolean)) => {
+            const openState = typeof value === "function" ? value(open) : value;
+            if (setOpenProp) {
+              setOpenProp(openState);
+            } else {
+              _setOpen(openState);
+            }
+
+            // This sets the cookie to keep the sidebar state.
+            document.cookie = `${SIDEBAR_COOKIE_NAME}=${openState}; path=/; max-age=${SIDEBAR_COOKIE_MAX_AGE}`;
```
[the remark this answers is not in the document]
dinmukhamedm: Good one, but this is auto-downloaded from shadcn

frontend/tailwind.config.js
dinmukhamedm, line 1: ouch, what's up with indents here?

dinmukhamedm (commented):
LGTM overall, left a few nitpicks, and there is clearly some more work to do

frontend/app/globals.css
adilb99: @dinmukhamedm that's one of my todos actually. These were the colors originally brought over from shadcn, but I ended up using the existing ones defined for the old navbar. So, we could either remove these or put them back into the classNames and tweak to be similar to the old navbar. The pros of the latter is we can decouple those colors from the navbar since they are also used elsewhere afaik (i.e. bg-primary, bg-secondary, etc.)

dinmukhamedm: @adilb99 Ah I see, you're saying we've already had these colors defined somewhere? Yeah, let's do classNames that we already have

frontend/assets/logo/laminar.svg
skull8888888: we already have new logo, check out how we do it in projects header and on landing header

skull8888888: so we should remove the old one

frontend/components/ui/header.tsx
```diff
@@ -22,8 +23,11 @@ export default function Header({ path, children, className }: HeaderProps) {
       className={`font-medium flex items-center justify-between flex-none h-12 border-b w-full ${className}`}
     >
       <div className="flex items-center">
+        {projectId.length > 0 && (
```
skull8888888: just add a prop, something like showSidebarButton and default to true, on /projects set it to false.

frontend/components/project/project-navbar.tsx
skull8888888: name and email underneath.

frontend/components/project/project-navbar.tsx
```diff
@@ -3,31 +3,50 @@
 import {
```
skull8888888, line 3: sidebar is too large, lots of empty space on the right to the text, let's make it smaller, around to the size of the largest text

skull8888888 (changes requested):

adilb99:
Addressed the comments. Gist:

- Better props in the header and avatar components
- Added email to the right of the avatar picture when expanded (username omitted due to being same as email)
- Narrower sidebar when expanded, aligned with the email
- Using new logo on expanded, deleted the old logo file + better naming for imported logos
- Decided to keep the sidebar css classes due to most of them still being used by sidebar.tsx

Let me know if there are any more improvements to be made. Otherwise, should be good to go!

skull8888888:
Looks good, few things.
- State of the sidebar should be saved. Seems like sidebar should already have this kind of functionality judging by the fact that it has some cookies stuff. If it doesn't let's just use localStorage
- padding on the avatar should be consistent with the rest of the sidebar, seems like sidebar has p-4 but avatar p-2?
Screenshot 2024-12-07 at 12 37 16 AM

- let's actually make max width of the sidebar to the width of the largest option, in this case it's playgrounds. Let's truncate email. Because aligning max size with email would cause sidebar to be very large if a user has a very long email.
- button should align with the rest of the content which has p-4. Align button so that it aligns with the rest of the content.
Screenshot 2024-12-07 at 12 42 26 AM

In general look at inconsistencies in the alignment, they immediately strike eyes.

adilb99:
Fixed the following:

- Smaller sidebar width closer aligned with Playgrounds menu item
- Fixed padding and alignment issues
- Sidebar state is saved in the cookies and properly used. (see the next comment for details)

frontend/components/ui/sidebar.tsx
```diff
@@ -70,10 +70,27 @@ const SidebarProvider = React.forwardRef<
...
+
+          if (sidebarCookie !== undefined) {
+            const savedState = sidebarCookie === 'true';
+            if (setOpenProp) {
+              setOpenProp(savedState);
+            } else {
+              _setOpen(savedState);
+            }
+          }
+        }, [setOpenProp]); // Only runs once on mount
```
adilb99: This is needed to handle SSR properly and not use `document` var before it is populated. However, this produces a slight lag when applying sidebar state on reload. Same occurs with proper usage of localStorage.

Possible solution is to retrieve the cookies on the server-side in layout.tsx and passing it into the component as a prop, but then propagating it back server-side to properly save makes the code a little too complex.

Possibly not a problem cause local frontend is a little slow and won't be so visible in production environment, but let me know your thoughts. @skull8888888 @dinmukhamedm

frontend/tailwind.config.js
```diff
@@ -9,84 +9,101 @@ module.exports = {
...
-        '3xl': ['1.875rem', { lineHeight: '2.25rem', letterSpacing: '-0.025em' }],
-        '4xl': ['2.25rem', { lineHeight: '2.5rem', letterSpacing: '-0.025em' }],
-        '5xl': ['3rem', { lineHeight: '1', letterSpacing: '-0.025em' }],
-        '6xl': ['3.75rem', { lineHeight: '1', letterSpacing: '-0.025em' }],
-        '7xl': ['4.5rem', { lineHeight: '1', letterSpacing: '-0.025em' }],
-        '8xl': ['6rem', { lineHeight: '1', letterSpacing: '-0.025em' }],
-        '9xl': ['8rem', { lineHeight: '1', letterSpacing: '-0.025em' }],
-      },
-    },
+  	container: {
```
dinmukhamedm: looks like indents are messed up again

dinmukhamedm (commented):
LGTM. Not explicitly approving as @skull8888888 has this checked out, so he'll have another look before we merge

frontend/tailwind.config.js
```diff
@@ -10,83 +10,92 @@ module.exports = {
...
+        }
       },
       animation: {
-        "accordion-down": "accordion-down 0.2s ease-out",
-        "accordion-up": "accordion-up 0.2s ease-out",
+        'accordion-down': 'accordion-down 0.2s ease-out',
+        'accordion-up': 'accordion-up 0.2s ease-out'
       },
       fontSize: {
-        'lg': ['1.125rem', { lineHeight: '1.6rem', letterSpacing: '-0.0125em' }],
```
skull8888888, line 78: bring 'lg' and 'xl' back and put them in comas for consistency

skull8888888 (changes requested):
last change, otherwise lgtm

PR #288, opened 2024-12-23
added datapoint count on dataset table

### LINKED ISSUE #274: Add "items count" column to the list of datasets
Add "items count" column to the list of datasets which would simply show how many datapoints given dataset has.

devansh-m12:
fixes #274 

@dinmukhamedm  added "DataPoints Count" in dataset table, ss for refrence let me know for any changes
Screenshot 2024-12-23 at 7 41 16 PM

----

> [!IMPORTANT]
> Add datapoints count to dataset table in both backend and frontend.
> 
>   - **Backend**:
>     - In `route.ts`, added `datapointsCount` to the dataset query using a subquery to count datapoints from `datasetDatapoints`.
>   - **Frontend**:
>     - In `datasets.tsx`, added a new column `datapointsCount` to the dataset table to display the count of datapoints.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=lmnr-ai%2Flmnr&utm_source=github&utm_medium=referral)<sup> for c4cf0bdd57a1bbbbbbda8231b49ae981d3691553. It will automatically update as commits are pushed.</sup>


dinmukhamedm (commented):
@devansh-m12 looks good overall, quick question: did you test it with empty datasets? By definition SQL's `COALESCE` may return null, because it returns first non-null value out of its arguments. Just to make sure things are working fine, we should test that case too.

Also, can you set the target branch to `dev`?

frontend/components/datasets/datasets.tsx
```diff
@@ -83,6 +83,11 @@ export default function Datasets() {
       header: 'name',
       size: 300
     },
+    {
+      accessorKey: 'datapointsCount',
```
dinmukhamedm, line 87: nit: update the dataset type (better introduce the new one) to include this column.

E.g. something like `type DatasetInfo = Dataset & {datapointsCount: number}`

dinmukhamedm (changes requested):
@devansh-m12 left a small nit. I'll merge this once it's fixed

dinmukhamedm (approved):

PR #302, opened 2025-01-05
api for uploading spans

dinmukhamedm:
> [!IMPORTANT]
> Adds an API endpoint for uploading spans to a queue, updates server configuration, and modifies TypeScript route handling for optional fields.
> 
>   - **API**:
>     - Adds `push_to_queue` function in `queues.rs` to handle POST requests to `/queues/push`.
>     - Defines `PushItem` and `PushToQueueRequest` structs for deserializing request data.
>     - Inserts spans into the database and pushes them to a labeling queue.
>   - **Server**:
>     - Registers `push_to_queue` service in `main.rs` under the `/v1` scope.
>   - **TypeScript**:
>     - Updates `remove/route.ts` to handle optional `action` fields with null checks.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=lmnr-ai%2Flmnr&utm_source=github&utm_medium=referral)<sup> for c0984e4a4bf787f96acf484594d55b6e899ef865. It will automatically update as commits are pushed.</sup>


app-server/src/api/v1/spans.rs
```diff
@@ -0,0 +1,130 @@
...
+#[derive(Deserialize)]
+#[serde(rename_all = "camelCase")]
+struct UploadSpansRequest {
+    spans: Vec<UploadSpan>,
+    #[serde(default)]
+    queue_name: Option<String>,
+}
+
+#[post("/spans/upload")]
+async fn upload_spans(
```
skull8888888: you shouldn't just upload spans, in fact you should never upload spans. Idea is that you push to the labeling queue from API, we just turn this pushed data into spans, and insert spans into queue

skull8888888 (changes requested):

app-server/src/api/v1/queues.rs
```diff
@@ -0,0 +1,111 @@
...
+
+#[derive(Deserialize)]
+#[serde(rename_all = "camelCase")]
+struct UploadToQueueRequest {
+    items: Vec<UploadItem>,
+    queue_name: String,
+}
+
+#[post("/queues/upload")]
+async fn upload_to_queue(
```
skull8888888: minor nit, but I would say push is more fitting than upload

app-server/src/api/v1/queues.rs
```diff
@@ -0,0 +1,111 @@
...
+        DB,
+    },
+    evaluations::utils::LabelingQueueEntry,
+    routes::types::ResponseResult,
+    traces::span_attributes::ASSOCIATION_PROPERTIES_PREFIX,
+};
+
+#[derive(Deserialize)]
+#[serde(rename_all = "camelCase")]
+struct UploadItem {
```
skull8888888: should rename here and below too

PR #347, opened 2025-01-22
add spans, datasets count to project card

olzhik11:
image

----

> [!IMPORTANT]
> Add datasets and spans count to project card, update API to fetch these counts, and introduce a pluralization utility.
> 
>   - **API Changes**:
>     - Update `GET` function in `route.ts` to include `datasetsCount` and `spansCount` for projects.
>     - Add `leftJoin` with `datasets` and `spans` tables and `groupBy` clause.
>   - **UI Changes**:
>     - Update `ProjectCard` component in `project-card.tsx` to display `datasetsCount` and `spansCount` using `pluralize` function.
>   - **Database**:
>     - Add index `datasets_project_id_hash_idx` in `0014_yellow_hannibal_king.sql` for `datasets` table.
>   - **Utilities**:
>     - Add `pluralize` function in `utils.ts` for handling singular/plural text.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=lmnr-ai%2Flmnr&utm_source=github&utm_medium=referral)<sup> for 5cb6ec46a9eff17817a2d5e1ee66753882235ca9. It will automatically update as commits are pushed.</sup>


frontend/app/api/workspaces/route.ts
```diff
@@ -46,9 +53,14 @@ export async function GET(req: NextRequest): Promise<Response> {
           id: projects.id,
           name: projects.name,
           workspaceId: projects.workspaceId,
+          datasetsCount: count(datasets.id),
+          spansCount: count(spans.spanId)
         })
         .from(projects)
-        .where(eq(projects.workspaceId, workspace.id));
+        .leftJoin(datasets, eq(projects.id, datasets.projectId))
+        .leftJoin(spans, eq(projects.id, spans.projectId))
```
dinmukhamedm: I'm afraid this might be slow. Let me run some `EXPLAIN ANALYZE`s against the prod DB and maybe either add an index or ask you to query from clickhouse instead

dinmukhamedm: Right, this is indeed fairly heavy. As a v0, can we add a separate api endpoint that, given project id, will return just the counts. This endpoint shall be queried from client side. This is so that all the data is loaded quickly and the heavy stats are populated separately later

frontend/components/projects/project-card.tsx
```diff
@@ -1,13 +1,20 @@
...
 import Link from 'next/link';
 
-import { Project } from '@/lib/workspaces/types';
+import { Skeleton } from '@/components/ui/skeleton';
+import { Project, ProjectStats } from '@/lib/workspaces/types';
 
 interface ProjectCardProps {
   project: Project;
+  stats?: ProjectStats;
+  isLoading: boolean;
```
dinmukhamedm: nit, and sort of out of scope of this PR. I prefer moving optional props to the end.

frontend/app/api/workspaces/route.ts
```diff
@@ -1,10 +1,16 @@
-import { desc, eq, sql } from 'drizzle-orm';
+import { desc, eq, sql} from 'drizzle-orm';
```
dinmukhamedm: nit: I don't think that's lint/prettier, probably need to add the space back after `sql`

dinmukhamedm (commented):
LGTM, but let's add another migration with index evaluations on project_id

frontend/components/projects/project-card.tsx
```diff
@@ -16,7 +24,42 @@ export default function ProjectCard({ project }: ProjectCardProps) {
...
+            {project.id}
+          </p>
+          <div className="flex gap-8">
+            {isLoading ? (
+              <>
+                <div className="flex flex-col gap-2">
+                  <Skeleton className="w-12 h-5" />
+                  <Skeleton className="w-8 h-6" />
+                </div>
+                <div className="flex flex-col gap-2">
```
dinmukhamedm, line 32: is this intentional that for skeletons the gap is 2 and for text it's 1? Would you mind attaching screenshots/recordings, so we can have a look?

frontend/lib/clickhouse/spans.ts
```diff
@@ -9,24 +9,23 @@ import {
...
   TimeRange
-} from "./utils";
+} from './utils';
 
 export enum SpanMetricGroupBy {
   Model = 'model',
   Provider = 'provider',
   Path = 'path',
-  Name = 'name',
+  Name = 'name'
```
dinmukhamedm: nit: let's agree on the trailing comma configuration in prettier. I prefer es6(?) don't remember the name, but the one that enables the trailing comma.

frontend/lib/clickhouse/spans.ts
```diff
@@ -151,7 +171,29 @@ export const getSpanMetricsSummary = async (
       projectId,
       nullValue: NULL_VALUE,
       types: [SpanType.DEFAULT, SpanType.LLM]
-    },
+    }
   });
   return await result.json();
 };
+
+export const getSpansCountInProject = async (
```
dinmukhamedm: nit: I think, unlike Rust, we don't need to pass the client from the route, because we can just import it directly in that file. I think this was my mistake initially in the spans-related functions. Could you check if this works without passing the client from the route?

dinmukhamedm (commented):
LGTM, left a few small questions/comments

frontend/lib/clickhouse/spans.ts
olzhik11: @dinmukhamedm I think my prettier was a bit off, I checked `package.json`, and found some prettier config there, I think we should remove it, i will include it in this PR if its fine

frontend/components/projects/project-card.tsx
olzhik11: @dinmukhamedm oh that one I should fix, I probably was testing, good catch

frontend/lib/clickhouse/spans.ts
olzhik11: @dinmukhamedm agree on that I was following a pattern that was used, I think it will work, will fix that

frontend/lib/clickhouse/spans.ts
dinmukhamedm: sure, go for it

olzhik11:
@dinmukhamedm all fixed, I also fixed other functions with clickhouse as param

dinmukhamedm (approved):
LGTM

PR #353, opened 2025-01-28
feat: add restricted access on github auth

olzhik11:
Added access control on `github` auth through json file

----

> [!IMPORTANT]
> Added restricted access for GitHub authentication using a JSON file to control allowed emails.
> 
>   - **Authentication**:
>     - Added `signIn` callback in `auth.ts` to restrict GitHub access based on emails from `allowed-emails.json`.
>     - Introduced `getEmailsConfig` in `server-utils.ts` to read allowed emails from JSON.
>   - **Components**:
>     - Updated `GitHubSignInButton` in `github-signin.tsx` to handle sign-in errors with toast notifications.
>     - Simplified `EmailSignInButton` in `email-signin.tsx` by removing unused props.
>   - **Features**:
>     - Added check for `GITHUB_AUTH` in `features.ts` to ensure GitHub credentials are set.
>   - **Misc**:
>     - Standardized import quotes across modified files.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=lmnr-ai%2Flmnr&utm_source=github&utm_medium=referral)<sup> for 855bacc158841175379b85d495ed0abe4bd7de6d. It will automatically update as commits are pushed.</sup>


frontend/components/sign-in/github-signin.tsx
```diff
@@ -1,11 +1,11 @@
-'use client';
+"use client";
 
-import { signIn } from 'next-auth/react';
-import * as React from 'react';
+import { signIn } from "next-auth/react";
+import * as React from "react";
```
dinmukhamedm: nit: redundant import

frontend/components/sign-in/github-signin.tsx
```diff
@@ -14,32 +14,31 @@ interface GitHubSignInButtonProps extends ButtonProps {
...
 }: GitHubSignInButtonProps) {
-  const [isLoading, setIsLoading] = React.useState(false);
+  const [isLoading, setIsLoading] = useState(false);
+
+  const handleSignIn = async () => {
+    try {
+      setIsLoading(true);
+      await signIn("github", { callbackUrl });
+    } catch (error) {
+      console.error(error);
```
dinmukhamedm: should we maybe do toast in addition to/instead of console.log?

frontend/lib/server-utils.ts
```diff
@@ -0,0 +1,21 @@
+import fs from "fs";
+import path from "path";
+
+export const getEmailsConfig = async (): Promise<string[] | false> => {
+  try {
+    const filePath = path.join(process.cwd(), "allowed-emails.json");
+
+    if (!fs.existsSync(filePath)) {
+      console.log("file not found :(");
```
dinmukhamedm: nit: do we really need this? If this is not debug log I suggest we either remove it or make it more informative

frontend/lib/server-utils.ts
```diff
@@ -0,0 +1,21 @@
...
+      return false;
+    }
+
+    const fileContent = fs.readFileSync(filePath, "utf-8");
+
+    const jsonData = JSON.parse(fileContent) as { emails: string[] };
+    return jsonData.emails;
+  } catch (e) {
+    console.error(e);
```
dinmukhamedm: nit: let's make this a little more informative, e.g. Failed to read allowed-emails.json ${e}

dinmukhamedm (commented):
Good job. I left a couple comments

frontend/lib/server-utils.ts
```diff
@@ -0,0 +1,21 @@
...
+
+    if (!fs.existsSync(filePath)) {
+      console.log("file not found :(");
+      return false;
+    }
+
+    const fileContent = fs.readFileSync(filePath, "utf-8");
+
+    const jsonData = JSON.parse(fileContent) as { emails: string[] };
+    return jsonData.emails;
```
dinmukhamedm: Even though this may throw TS errors, I think it will silently return `undefined` for valid JSON files with invalid format. Please handle, by at least returning `jsonData.emails ?? []` or better throw an error if `jsonData.emails === undefined`

dinmukhamedm: Even though this may throw TS errors, I think it will silently return `undefined` for valid JSON files with invalid format. Please handle, by at least returning `jsonData.emails ?? []` or better throw an error if `jsonData.emails === undefined`

dinmukhamedm: + we might also want to validate every email, but that might be overkill – up to you

olzhik11:
@dinmukhamedm yes agree with all comments, should have fixed them before, should be good now, for the emails part I agree we need fallback or error throw, I added error throw though if needed we can make fallback for empty list, meaning no user will be able to pass github signin

frontend/lib/server-utils.ts
olzhik11: I think email validation might be too much

dinmukhamedm (dismissed):
LGTM

frontend/lib/server-utils.ts
```diff
@@ -0,0 +1,22 @@
...
+export const getEmailsConfig = async (): Promise<string[] | false> => {
+  const filePath = path.join(process.cwd(), allowedEmailsFileName);
+
+  if (!fs.existsSync(filePath)) {
+    return false;
+  }
+
+  const fileContent = fs.readFileSync(filePath, "utf-8");
+
+  const jsonData = JSON.parse(fileContent) as { emails?: string[] };
```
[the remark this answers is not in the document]
dinmukhamedm: @olzhik11 please address

frontend/lib/server-utils.ts
```diff
@@ -0,0 +1,22 @@
...
+const allowedEmailsFileName = "allowed-emails.json";
+
+export const getEmailsConfig = async (): Promise<string[] | false> => {
+  const filePath = path.join(process.cwd(), allowedEmailsFileName);
+
+  if (!fs.existsSync(filePath)) {
+    return false;
+  }
+
+  const fileContent = fs.readFileSync(filePath, "utf-8");
```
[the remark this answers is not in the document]
dinmukhamedm: @olzhik11 please have a look if that makes sense

frontend/lib/server-utils.ts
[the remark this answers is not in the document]
olzhik11: @dinmukhamedm for this one then we won't be able to throw error if file doesn't match structure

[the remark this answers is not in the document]
olzhik11: @dinmukhamedm I added fallback instead, and throw on file parsing error

dinmukhamedm (approved):

Issue #345, opened 2025-01-22
How to set auth and where to set the Laminar api key?


enniofilicicchia:
I am trying this as an alternative to Langfuse that is slow, but I cannot find informations about remote hosting in the code nor in the docs.

How to set an authentication method? I don't want everyone to be able to create an account on my laminar
What is the api key? is it NEXT_AUTH or SHARED_SECRET?

To my understanding self-hosting does support only local hosting so far (no other mentions in the docs).

skull8888888:
@enniofilicicchia thanks for the issue! Yep, right now auth is very basic in the self-hosted configuration. We're gonna release better auth very very soon. In the meantime, would love to hear your thoughts on how do you want to pre-populate users who has access to your managed version of Laminar?

We're working on guides on how to deploy Laminar properly in the cloud. What are you preferred options? AWS cloud formation, HELM chart, or Terraform?

enniofilicicchia:
Thank you for the answer

For both questions, I don't care much
I just need a page with a basic login + ability to set 1 api_key

skull8888888:
@enniofilicicchia In this case you already have it, just spin up the local instance, enter your email. Finish onboarding and generate local API key in the project page.

enniofilicicchia:
Ok thank you,
I just don't want everyone on the web to be able to add users, but for now it's fine

Issue #344, opened 2025-01-21
Production URL issue


DarkNeedle:
Just a recommendation, the documentation needs to be updated to allow this to be run against a valid URL vs localhost. The .env variable should be called up specifically. In the docker-compose-full.yml when adjusting the URL for a domain https://<subdomain>.<domain>.<tld> you will get a Page not found immediately after using any email address to login.

Merely setting NEXTAUTH_URL and NEXT_PUBLIC_URL should be sufficient

In setting a proper URL other than localhost which would be standard for hosting the error pops up in the URL
https://<some domain>/not-found?callbackUrl=%2Fproject%2F665df17d-a7d5-4318-8131-c16928fc6c77%2Ftraces

In middleware.ts, there's a configuration conflict:
Line 46: pages.signIn: '/not-found' - This is redirecting all unauthorized requests to /not-found
This conflicts with auth.ts which has pages.signIn: '/sign-in'
The middleware is making an API call to ${process.env.NEXTAUTH_URL}/api/auth to verify authorization, but this URL needs to be properly configured for production.

The auth flow appears to be:
User tries to access a protected route (/project/...)
Middleware checks authorization by calling /api/auth
If unauthorized, it redirects to /not-found instead of /sign-in
After login, it tries to redirect back but fails due to the middleware configuration

I have not performed a full code analysis yet but this is a fundamental issue that should have been caught. Likewise, why not enable the full Auth flow instead of processing just an email address. 

-DB

skull8888888:
@DarkNeedle Thanks for the issue! Totally valid points, in fact we're working on improving auth experience in the self-hosted version right now. Stay tuned. Will release ASAP!

DarkNeedle:
Thanks. Is there an ETA or where this sits with regard to priority. I can go in and address it myself in the code but if this is something already in the works and roadmapped I'd prefer to hold off, unless it's months out.

-DB

skull8888888:
Should be couple of days, although still thinking of a ways to properly structure auth when people deploy their own instances of Laminar. Do you have any thoughts on a better approach regarding user registration? Because I can imagine when someone deploys their own instance, people only want known users to use the platform. I'm now thinking whether we should pre-populate the database with existing emails or something like this?

Interested in your thoughts since I assume you want to deploy your own version.


> The middleware is making an API call to ${process.env.NEXTAUTH_URL}/api/auth to verify authorization, but this URL needs to be properly configured for production.

also regarding this. Can you elaborate on that?

[...]

PR #370, opened 2025-02-03
fix: fix github sign-in button

olzhik11:
Fixed github signin button

image


dinmukhamedm (approved):
LGTM. @skull8888888 should we also notify people in #344 and #345 and close their issues if they don't have further questions?

PR #397, opened 2025-02-15
redis (valkey) cache

skull8888888:
> [!IMPORTANT]
> This pull request replaces the in-memory cache with a Redis-based caching system, updating both backend and frontend components to integrate the new cache interface.
> 
>   - **Cache Implementation**:
>     - Introduces `RedisCache` in `cache/redis.rs` and `InMemoryCache` in `cache/in_memory.rs`.
>     - Updates `CacheTrait` in `cache/mod.rs` to support both Redis and in-memory caching.
>     - Adds cache key constants in `cache/keys.rs`.
>   - **Backend Integration**:
>     - Replaces in-memory cache with Redis in `main.rs`, using `REDIS_URL` from environment.
>     - Updates cache usage in `api/utils.rs`, `auth/mod.rs`, `routes/api_keys.rs`, and `routes/pipelines.rs` to use new cache interface.
>     - Removes old cache implementation in `cache/cache.rs`.
>   - **Frontend Integration**:
>     - Adds `cache.ts` for frontend caching with Redis support.
>     - Updates `api/projects/route.ts` to use new cache system.
>     - Modifies `db/utils.ts` to cache user membership checks.
>   - **Miscellaneous**:
>     - Updates `.env.example` files to include `REDIS_URL`.
>     - Removes unused code and refactors existing code for consistency with new cache system.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=lmnr-ai%2Flmnr&utm_source=github&utm_medium=referral)<sup> for 7768fe978d76cb91558983ff4bad2cfc7eceb1ba. It will automatically update as commits are pushed.</sup>


app-server/src/cache/redis.rs
```diff
@@ -0,0 +1,81 @@
...
+use super::cache::CacheTrait;
+
+pub struct RedisCache<T> {
+    connection: MultiplexedConnection,
+    _phantom: PhantomData<T>,
+}
+
+impl<T> RedisCache<T>
+where
+    T: Serialize + DeserializeOwned + Send + Sync + 'static,
```
dinmukhamedm: nit: any reason why we use `DeserializeOwned` over `for<'de> Deserialize<'de>`? Reference: https://stackoverflow.com/questions/35592750/how-does-for-syntax-differ-from-a-regular-lifetime-bound/

app-server/src/cache/redis.rs
```diff
@@ -0,0 +1,81 @@
...
+    T: Serialize + DeserializeOwned + Send + Sync + Clone + 'static,
+{
+    async fn get(&self, key: &str) -> Option<Box<dyn Any>> {
+        let result: RedisResult<Vec<u8>> = self.connection.clone().get(key).await;
+        match result {
+            Ok(bytes) => {
+                if bytes.is_empty() {
+                    return None;
+                }
+                match bincode::deserialize::<T>(&bytes) {
```
dinmukhamedm: genuinely curious: what are the advantages of using a separate crate over something like:

```
serde_json::from_str::<T>(&String::from_utf8(&bytes));
```

or even:

```
serde_json::from_slice::<T>(&bytes);
```

and similarly in serialize?

I suppose this is to do with efficiency and/or configurability, but (a) is it more efficient now with default options, (b) will we need this configurability any time soon?

dinmukhamedm (commented):
Looks good overall, left a couple questions. Major questions/suggestions to this PR:

1. Why is it only the project api key that we are moving? Gradual migration, WIP, or is it intended so?
2. Let's move the code around in the `cache` module for clarity and consistency with the other traits. I'd probably move most of what is in `src/cache/cache.rs` now to `src/cache/mod.rs`, rename the former to something like `moka_cache.rs` or `in_memory.rs`. Finally, maybe even have a struct `InMemoryCache<T>` that would hold one field `MokaCache<String, T>` inside.
3. The last point in two above would probably also allow us to get rid of most, if not all, scary `Any`s and `downcast`s, but this might be outside of scope of this PR

app-server/src/cache/mod.rs
```diff
@@ -1,2 +1,36 @@
...
+    Redis(RedisCache),
+}
+
+#[async_trait]
+#[enum_dispatch(Cache)]
+pub trait CacheTrait: Sync + Send {
+    async fn get<T>(&self, key: &str) -> Result<Option<T>, CacheError>
+    where
+        T: for<'de> Deserialize<'de> + Send + Sync + 'static;
+    async fn insert<T>(&self, key: &str, value: T) -> Result<(), CacheError>
```
skull8888888, line 32: Do we need generic on insert and remove?

frontend/lib/db/utils.ts
dinmukhamedm, line 1: @skull8888888 I left this file in `lib/db/` but let me know if you think I should move this somewhere else, like we did in app-server, i.e. there is a function that calls cache and calls db if needed as a wrapper around this function

PR #423, opened 2025-03-02
anthropic cache

skull8888888:
> [!IMPORTANT]
> Introduces caching for input tokens in Anthropic provider, updating cost estimation and token handling logic.
> 
>   - **Behavior**:
>     - Introduces `estimate_input_cost` and `estimate_output_cost` functions in `costs/mod.rs` to handle input and output token cost estimation separately.
>     - Adds caching logic for input tokens in `estimate_input_cost`, considering `regular_input_tokens`, `cache_write_tokens`, and `cache_read_tokens`.
>     - Updates `get_llm_usage_for_span` in `utils.rs` to use the new input token structure.
>   - **Data Structures**:
>     - Modifies `InputTokens` struct in `spans.rs` to include `regular_input_tokens`, `cache_write_tokens`, and `cache_read_tokens`.
>     - Updates `SpanAttributes::input_tokens()` to return `InputTokens` with cache-related fields.
>   - **Misc**:
>     - Removes `TokensKind` enum from `costs/mod.rs`.
>     - Changes token type from `u32` to `i64` in `calculate_cost` in `providers/utils.rs` and related functions.
>     - Adds constants `GEN_AI_CACHE_WRITE_INPUT_TOKENS` and `GEN_AI_CACHE_READ_INPUT_TOKENS` in `span_attributes.rs`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=lmnr-ai%2Flmnr&utm_source=github&utm_medium=referral)<sup> for d8bfd7eabdbd21c89bc3f282a853681e118946a4. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/app-server/src/traces/span_attributes.rs
+++ b/app-server/src/traces/span_attributes.rs
@@ -30,3 +30,7 @@ pub const SPAN_TYPE: &str = "lmnr.span.type";
 pub const SPAN_PATH: &str = "lmnr.span.path";
 pub const SPAN_IDS_PATH: &str = "lmnr.span.ids_path";
 pub const LLM_NODE_RENDERED_PROMPT: &str = "lmnr.span.prompt";
+
+pub const GEN_AI_ANTHROPIC_CACHE_WRITE_INPUT_TOKENS: &str =
+    "gen_ai.usage.cache_creation_input_tokens";
+pub const GEN_AI_ANTHROPIC_CACHE_READ_INPUT_TOKENS: &str = "gen_ai.usage.cache_read_input_tokens";
--- a/app-server/src/traces/span_attributes.rs
+++ b/app-server/src/traces/span_attributes.rs
@@ -31,6 +31,5 @@ pub const SPAN_PATH: &str = "lmnr.span.path";
 pub const SPAN_IDS_PATH: &str = "lmnr.span.ids_path";
 pub const LLM_NODE_RENDERED_PROMPT: &str = "lmnr.span.prompt";
 
-pub const GEN_AI_ANTHROPIC_CACHE_WRITE_INPUT_TOKENS: &str =
-    "gen_ai.usage.cache_creation_input_tokens";
-pub const GEN_AI_ANTHROPIC_CACHE_READ_INPUT_TOKENS: &str = "gen_ai.usage.cache_read_input_tokens";
+pub const GEN_AI_CACHE_WRITE_INPUT_TOKENS: &str = "gen_ai.usage.cache_creation_input_tokens";
+pub const GEN_AI_CACHE_READ_INPUT_TOKENS: &str = "gen_ai.usage.cache_read_input_tokens";
--- a/app-server/src/traces/spans.rs
+++ b/app-server/src/traces/spans.rs
@@ -105,7 +105,7 @@ impl SpanAttributes {
     }
 
     pub fn input_tokens(&mut self) -> InputTokens {
-        let regular_input_tokens =
+        let total_input_tokens =
             if let Some(Value::Number(n)) = self.attributes.get(GEN_AI_INPUT_TOKENS) {
                 n.as_i64().unwrap_or(0)
             } else if let Some(Value::Number(n)) = self.attributes.get(GEN_AI_PROMPT_TOKENS) {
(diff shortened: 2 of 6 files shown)


app-server/src/traces/span_attributes.rs
```diff
@@ -30,3 +30,7 @@ pub const SPAN_TYPE: &str = "lmnr.span.type";
 pub const SPAN_PATH: &str = "lmnr.span.path";
 pub const SPAN_IDS_PATH: &str = "lmnr.span.ids_path";
 pub const LLM_NODE_RENDERED_PROMPT: &str = "lmnr.span.prompt";
+
+pub const GEN_AI_ANTHROPIC_CACHE_WRITE_INPUT_TOKENS: &str =
+    "gen_ai.usage.cache_creation_input_tokens";
+pub const GEN_AI_ANTHROPIC_CACHE_READ_INPUT_TOKENS: &str = "gen_ai.usage.cache_read_input_tokens";
```
dinmukhamedm: nit, non-blocking: I think we decided with nirga that this is not named anthropic for that some other providers charge using similar pricing model. So we could remove "anthropic" from variable names

app-server/src/traces/spans.rs
```diff
@@ -90,18 +106,47 @@ impl SpanAttributes {
...
+                // updating to the new convention
+                let n = n.as_i64().unwrap_or(0);
+                self.attributes
+                    .insert(GEN_AI_INPUT_TOKENS.to_string(), json!(n));
+                n
+            } else {
+                0
+            };
+
+        if self.provider_name() == Some("anthropic".to_string()) {
```
dinmukhamedm, line 121: nit (again non-blocking) we may want to test this matching with opentelemetry-instrumentation-langchain with langchain-anthropic

app-server/src/traces/spans.rs
```diff
@@ -90,18 +106,47 @@ impl SpanAttributes {
...
-                .insert(GEN_AI_INPUT_TOKENS.to_string(), json!(n));
-            n
+    pub fn input_tokens(&mut self) -> InputTokens {
+        let mut input_tokens = InputTokens {
+            regular_input_tokens: 0,
+            cache_write_tokens: 0,
+            cache_read_tokens: 0,
+        };
+
+        let regular_input_tokens =
```
dinmukhamedm: I'd probably rename this variable to something like `total_input_tokens`, because regardless if there is cache `gen_ai.usage.input_tokens` (or `gen_ai.usage.prompt_tokens`) is meant to represent the total number of input tokens

app-server/src/traces/spans.rs
```diff
@@ -90,18 +106,47 @@ impl SpanAttributes {
...
-            let n = n.as_i64().unwrap_or(0);
-            self.attributes
-                .insert(GEN_AI_INPUT_TOKENS.to_string(), json!(n));
-            n
+    pub fn input_tokens(&mut self) -> InputTokens {
+        let mut input_tokens = InputTokens {
+            regular_input_tokens: 0,
+            cache_write_tokens: 0,
+            cache_read_tokens: 0,
+        };
```
dinmukhamedm: I'm probably being too picky, but I'd prefer if this was immutable for readability and more idiomatic rust. I.e. let's construct the struct at return (in each of if/else)

app-server/src/traces/spans.rs
```diff
@@ -152,6 +197,12 @@ impl SpanAttributes {
             }
         } else {
             None
+        };
+
+        if let Some(name) = name {
+            Some(name.to_lowercase().trim().to_string())
+        } else {
+            None
```
dinmukhamedm: super nit: can do with `name.map(|name| name.to_lowercase()...)`

dinmukhamedm (commented):
Many nits, one major comment: I'd need to check what we do for openai on the instrumentation side, and see if we need to match against `anthropic` provider name in this PR at all

app-server/src/traces/spans.rs
skull8888888: it's meant to convey that this token will be priced as regular ones, hence the name

dinmukhamedm: Doesn't make sense to me. Below in line 141 we subtract the cached tokens from this variable. So my understanding is: extract _total_ input tokens from attributes, then
1. if the request has cache tokens, subtract the cache tokens to get _regular_ tokens
2. otherwise set the _regular_ token count to the _total_ token count

PR #427, opened 2025-03-05
feat: labels v1 (2)

olzhik11:
- removed labels value, and label-classes value-map fields, functionalities
- added color, and db constraints to uniqueness of label-classes
- added new ui to add/remove labels from span
- added dashboard chart for labels frequency
- made appropriate fixes on app-server, and db/clickhouse migrations


dinmukhamedm (approved):

docker-compose-local-dev.yml
```diff
@@ -25,18 +25,23 @@ services:
       retries: 5
 
   clickhouse:
-    build:
-      context: ./clickhouse
+    image: clickhouse/clickhouse-server:latest
```
dinmukhamedm, line 28: @olzhik11 why did we remove ports in this file?

olzhik11: will add back, they got deleted

PR #485, opened 2025-04-11
readme

skull8888888:


frontend/components/landing/pricing.tsx
dinmukhamedm, line 1: sliders and calculation for steps? Also sliders for hobby tier?

PR #500, opened 2025-04-15
feat: add real time update on sessions LAM-532

olzhik11:
- real time updates on sessions in chat sidebar
- supabase client with .evn variables
- migration on `agent_chats` table
- update `agent_chats` table in agent worker in app-server

----

> [!IMPORTANT]
> Add real-time updates for chat sessions using Supabase and update database schema and frontend components accordingly.
> 
>   - **Real-time Updates**:
>     - Implement real-time updates for chat sessions in `AgentSidebar` using Supabase.
>     - Subscribe to `agent_chats` table changes and update session data accordingly.
>   - **Database**:
>     - Add `agent_status` column to `agent_chats` table with default 'idle' in `0028_futuristic_blonde_phantom.sql`.
>     - Update `update_agent_chat_status()` in `agent_chats.rs` to modify `agent_status`.
>   - **Frontend**:
>     - Modify `AgentSidebar` to use `supabaseClient` for real-time updates.
>     - Update `route.ts` to include `agentStatus` in session data.
>     - Add `supabaseAccessToken` to `ChatUser` in `layout.tsx` and `pricing-context.tsx`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=lmnr-ai%2Flmnr&utm_source=github&utm_medium=referral)<sup> for c44ebd89039921a2e6400f1600fbdc637d597be8. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/app-server/src/db/agent_chats.rs
+++ b/app-server/src/db/agent_chats.rs
@@ -1,7 +1,7 @@
 use sqlx::PgPool;
 use uuid::Uuid;
 
-pub async fn update_agent_chat(
+pub async fn update_agent_chat_status(
     pool: &PgPool,
     agent_status: &str,
     updated_at: chrono::DateTime<chrono::Utc>,
--- a/app-server/src/db/agent_chats.rs
+++ b/app-server/src/db/agent_chats.rs
@@ -1,0 +1,22 @@
+use sqlx::PgPool;
+use uuid::Uuid;
+
+pub async fn update_agent_chat(
+    pool: &PgPool,
+    agent_status: &str,
+    updated_at: chrono::DateTime<chrono::Utc>,
+    session_id: &Uuid,
+) -> anyhow::Result<()> {
+    sqlx::query(
+        "UPDATE agent_chats
+        SET agent_status = $1, updated_at = $2
+        WHERE session_id = $3",
+    )
+    .bind(agent_status)
+    .bind(updated_at)
+    .bind(session_id)
+    .execute(pool)
+    .await?;
+
+    Ok(())
+}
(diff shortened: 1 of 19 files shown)


app-server/src/db/agent_chats.rs
```diff
@@ -0,0 +1,22 @@
+use sqlx::PgPool;
+use uuid::Uuid;
+
+pub async fn update_agent_chat(
```
skull8888888: let's make more specific, update_agent_chat_status

frontend/lib/supabase.ts
```diff
@@ -0,0 +1,13 @@
+import { createClient } from "@supabase/supabase-js";
+
+export const createSupabaseClient = (accessToken: string) =>
+  createClient(process.env.NEXT_PUBLIC_SUPABASE_URL!, process.env.NEXT_PUBLIC_SUPABASE_ANON_KEY!, {
```
skull8888888: I think we already have supabase client somewhere defined. Also, NEXT_PUBLIC won't work

skull8888888 (changes requested):

frontend/lib/supabase.ts
dinmukhamedm: We should re-use the code here or at least make something similar: https://github.com/lmnr-ai/lmnr/blob/main/frontend/contexts/user-context.tsx

olzhik11: decided to keep this file for now, just replace .env variables, I will later make use of this function in other places

PR #586, opened 2025-05-30
show user id in traces view, allow filtering on it

dinmukhamedm:


frontend/components/traces/traces-table/index.tsx
```diff
@@ -98,6 +98,7 @@ export default function TracesTable({ traceId, onRowClick }: TracesTableProps) {
       }
 
       const data = (await res.json()) as PaginatedResponse<Trace>;
+      console.log(data.items);
```
skull8888888: caught before bots

dinmukhamedm: fastest in the wild west

olzhik11 (approved):

PR #713, opened 2025-07-16
feat: traces API LAM-744

olzhik11:
- /traces search endpoint
- /traces/:trace_id endpoint

----

> [!IMPORTANT]
> Add new API endpoints for trace search and retrieval, with enhanced filtering and error handling.
> 
>   - **API Endpoints**:
>     - Add `/traces/query` endpoint in `traces.rs` for searching traces with filters and pagination.
>     - Add `/traces/{trace_id}` endpoint in `traces.rs` for retrieving a specific trace by ID.
>   - **Database**:
>     - Implement `get_traces` and `get_trace` functions in `traces.rs` and `trace.rs` for database interactions.
>     - Add `filters.rs` for handling query filters, supporting various data types and operators.
>   - **Error Handling**:
>     - Extend `Error` enum in `error.rs` to include `BadRequest` for handling validation errors.
>   - **Misc**:
>     - Update `main.rs` to register new API endpoints.
>     - Add `traces` module to `ch/mod.rs` and `db/mod.rs`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=lmnr-ai%2Flmnr&utm_source=github&utm_medium=referral)<sup> for 2288b529f5d812cc8140be5049c6020efeee3fcf. You can [customize](https://app.ellipsis.dev/lmnr-ai/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


app-server/src/api/v1/mod.rs
```diff
@@ -5,4 +5,5 @@ pub mod evals;
 pub mod evaluators;
 pub mod metrics;
 pub mod tag;
+pub mod trace;
```
dinmukhamedm: no need to create a new file, we should keep the code in the same traces module

app-server/src/api/v1/traces.rs
```diff
@@ -66,3 +66,32 @@ pub async fn process_traces(
...
+
+
+#[derive(Serialize)]
+struct SearchTracesResponse {
+    data: Vec<traces::TraceInfo>,
+    count: i64
+}
+
+#[get("/traces")]
+pub async fn search_traces(
```
dinmukhamedm: nit: this isn't really search, is it? Because search query is optional. Probably `get_traces`

app-server/src/db/traces.rs
```diff
@@ -0,0 +1,388 @@
...
+    pub session_id: Option<String>,
+    pub input_token_count: i64,
+    pub output_token_count: i64,
+    pub total_token_count: i64,
+    pub input_cost: f64,
+    pub output_cost: f64,
+    pub cost: f64,
+    pub trace_type: TraceType,
+    pub status: Option<String>,
+    pub latency: f64,
```
dinmukhamedm: probably `duration` would be better (would have to rename the alias in the query).

app-server/src/db/traces.rs
```diff
@@ -0,0 +1,388 @@
...
+    pub metadata: Option<Value>,
+}
+
+#[derive(Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct SearchTracesParams {
+    page_size: Option<i32>,
+    page_number: Option<i32>,
```
dinmukhamedm: No way these are signed integers. Let's replace with u32 or usize or something and cast right before adding to query with `as i32`.

also nit: possibly `offset_pages`? We count from 0, and `page_number` sort of hints at counting from 1. No strong opinions though.

app-server/src/db/traces.rs
```diff
@@ -0,0 +1,388 @@
...
+    pub fn page_number(&self) -> i32 {
+        self.page_number.unwrap_or(0)
+    }
+    
+    pub fn offset(&self) -> i32 {
+        self.page_number() * self.page_size()
+    }
+
+    pub fn start_time(&self) -> DateTime<Utc> {
+        self.start_time.unwrap_or_else(|| Utc::now() - Duration::hours(24))
```
dinmukhamedm: nit: https://x.com/din_mlb/status/1945869032596369883
same below

app-server/src/db/traces.rs
```diff
@@ -0,0 +1,388 @@
...
+            parsed
+        }
+    }
+
+    pub fn filters(&self) -> &[Filter] {
+        &self.filters
+    }
+
+    pub fn validate_and_convert_filters(&mut self) -> Result<(), Error> {
+        let field_configs = create_traces_field_configs();
```
dinmukhamedm: must be a static const or something similar. Creating the same `HashMap` on every request is too expensive

app-server/src/db/traces.rs
```diff
@@ -0,0 +1,388 @@
...
+                    time_range.num_days()
+                )
+            ));
+        }
+        
+        Ok(())
+    }
+
+        fn validate_page_size(&self) -> Result<(), Error> {
+        let page_size = self.page_size();
```
dinmukhamedm: nit: indent

app-server/src/db/traces.rs
```diff
@@ -0,0 +1,388 @@
...
+        } else {
+            Err("Status filter only supports 'error' or 'success' values".to_string())
+        }
+    } else {
+        Err("Status must be a string".to_string())
+    }
+}
+
+
+fn build_trace_filters<'a>(
```
dinmukhamedm: super nit: let's rename the lifetime to q or qb to indicate that all parameters and the response live as long as the query builder

app-server/src/db/traces.rs
```diff
@@ -0,0 +1,388 @@
...
+    mut query_builder: QueryBuilder<'a, sqlx::Postgres>,
+    project_id: Uuid,
+    trace_ids: &'a Option<HashSet<Uuid>>,
```
dinmukhamedm: nit: trace_ids can be moved here, so we can do `into_iter` and avoid cloning. I know it's lightweight, still let's try to do so

app-server/src/db/traces.rs
```diff
@@ -0,0 +1,388 @@
...
+
+    let field_configs = create_traces_field_configs();
+    for filter in params.filters() {
+        if filter.field == "status" && filter.operator == FilterOperator::Eq {
+            if let FilterValue::String(status_value) = &filter.value {
+                if status_value == "success" {
+                    query_builder.push(" AND t.status IS NULL");
+                    continue;
+                }
+            }
```
dinmukhamedm, line 301: will this work if the status in the table is explicitly `"success"`? I know we don't do so now, but if we decide to change it, we'll most certainly forget to update here

also nit: let's add a tiny comment explaining what's going on here.

app-server/src/db/filters.rs
```diff
@@ -0,0 +1,423 @@
...
+            FilterOperator::NotILike => "NOT ILIKE",
+        }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[serde(untagged)]
+pub enum FilterValue {
+    String(String),
+    Number(f64),
```
dinmukhamedm: these are the actual values that are passed to the database? Why do we treat floats and integers the same way? I think it's cleaner and safer to actually have `Float(f64)` and `Integer(i64)`.

app-server/src/db/filters.rs
```diff
@@ -0,0 +1,423 @@
...
+                // Parse key=value format
+                if let Some(eq_pos) = s.find('=') {
+                    let key = s[..eq_pos].to_string();
+                    let value = s[eq_pos + 1..].to_string();
+                    Ok(FilterValue::Json { key, value })
+                } else {
+                    Err("JSON value must be in 'key=value' format".to_string())
+                }
+            },
+        }
```
dinmukhamedm: Can you paste an example of what the query should look like for this to work? Is it this?

```json
{
  "column": "metadata",
  "operator": "eq",
  "value": "environment=production"
}
```

What happens if the `=` sign is the last char, does this not panic?

I think this is too confusing, do you think adding a separate filter op for JSON and then some nested value must be passed only will make the code even messier? E.g.

```jsonc
{
  "column": "metadata",
  "operator": "eq",
  "value": {
     "key": "environment",
     "operator": "eq",  // only eq is supported for now
     "value": "production"
   }
}
```

Will we also need cast types here if the value is for example number?

app-server/src/db/filters.rs
```diff
@@ -0,0 +1,423 @@
...
+            (FilterOperator::ILike | FilterOperator::NotILike, FilterValue::String(s)) => {
+                query_builder.push(self.operator.to_sql());
+                query_builder.push(" ");
+                query_builder.push_bind(format!("%{}%", s));
+            },
+            (FilterOperator::Eq, FilterValue::Json { key, value }) => {
+                query_builder.push("@> ");
+                let json_obj = serde_json::json!({ key: value });
+                query_builder.push_bind(json_obj);
+            },
```
dinmukhamedm: we might want to re-write this using `column ->> key`. Let's discuss offline

app-server/src/db/filters.rs
```diff
@@ -0,0 +1,423 @@
...
+    Ok(filters)
+}
+
+struct ParsedField<'a> {
+    config: &'a FieldConfig,
+    field: String,
+    json_key: Option<String>,
+}
+
+fn parse_field<'a>(
```
dinmukhamedm, line 277: ok, I got to here and now I'm totally lost about how JSON filters must be formatted and which code paths do they go through. Can you explain offline?

app-server/src/db/filters.rs
```diff
@@ -0,0 +1,423 @@
...
+        }
+
+        if let Some(validator) = field_config.validator {
+            validator(&self.value)?;
+        }
+
+        Ok(())
+    }
+
+    pub fn apply_to_query_builder<'a>(
```
dinmukhamedm: ditto nit about lifetime name

app-server/src/db/filters.rs
```diff
@@ -0,0 +1,423 @@
...
+
+use serde::{Deserialize, Serialize};
+use sqlx::QueryBuilder;
+use uuid::Uuid;
+
+use crate::routes::error::Error;
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub enum FilterOperator {
+    #[serde(rename = "eq")]
```
dinmukhamedm: how about we allow the signs as well using `serde(alias)`?

app-server/src/db/traces.rs
```diff
@@ -0,0 +1,388 @@
...
+            t.session_id,
+            t.input_token_count, 
+            t.output_token_count, 
+            t.total_token_count,
+            t.input_cost, 
+            t.output_cost, 
+            t.cost, 
+            t.trace_type, 
+            t.status,
+            CAST(EXTRACT(EPOCH FROM (t.end_time - t.start_time)) * 1000 AS FLOAT8) as latency,
```
dinmukhamedm: if we rename this to duration, we'd have to rename it here

app-server/src/db/traces.rs
```diff
@@ -0,0 +1,388 @@
...
+    }
+
+    let count_query_builder = QueryBuilder::new(
+        "SELECT COUNT(t.id) as count 
+         FROM traces t
+         WHERE t.start_time IS NOT NULL AND t.end_time IS NOT NULL AND t.project_id = "
+    );
+    let mut count_query_builder = build_trace_filters(count_query_builder, project_id, &trace_ids, params)?;
+
+    let count_result: (i64,) = count_query_builder
```
dinmukhamedm: any idea why it is a tuple of `i64` not just a single one?

app-server/src/db/filters.rs
```diff
@@ -0,0 +1,423 @@
...
+        if let Some(validator) = field_config.validator {
+            validator(&self.value)?;
+        }
+
+        Ok(())
+    }
+
+    pub fn apply_to_query_builder<'a>(
+        &self,
+        mut query_builder: QueryBuilder<'a, sqlx::Postgres>,
```
dinmukhamedm: another super nit for both this function and `build_trace_filters`. Can we take a `&mut` reference to query builder and mutate in place? It's a bit confusing that we create a non-mutable query builder, then move it to this function, then it returns a mutable query builder which is logically the same query builder.

Might actually make the lifetime annotations slightly simpler even

app-server/src/db/filters.rs
```diff
@@ -0,0 +1,423 @@
...
+            }
+
+            let field = parts[0].to_string();
+            let operator = FilterOperator::from_str(parts[1])
+                .map_err(|e| Error::custom(e))?;
+            let value = FilterValue::String(parts[2].to_string());
+
+            filters.push(Filter::new(field, operator, value));
+        }
+    }
```
dinmukhamedm: This isn't reliable at all. Let's discuss offline what we can do instead

app-server/src/ch/traces.rs
```diff
@@ -0,0 +1,72 @@
...
+    trace_id: Uuid,
+}
+
+pub async fn search_spans_for_trace_ids(
+    clickhouse: &clickhouse::Client,
+    project_id: Uuid,
+    search_query: &str,
+    params: &GetTracesParams,
+) -> Result<Option<HashSet<Uuid>>, Box<dyn std::error::Error + Send + Sync>> {
+    let search_fields = params.search_in();
```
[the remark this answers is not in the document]
olzhik11: handled on impl of params

PR #715, opened 2025-07-16
batch spans for write into DB

dinmukhamedm:
> [!IMPORTANT]
> Enhances span processing by implementing batch operations for database writes and message queue handling, improving performance and scalability.
> 
>   - **Batch Processing**:
>     - Introduces `insert_spans_batch()` in `ch/spans.rs` for batch insertion of spans into Clickhouse.
>     - Implements `record_spans_batch()` in `db/spans.rs` for batch recording of spans into Postgres.
>     - Updates `update_trace_attributes_batch()` in `db/trace.rs` for batch updating trace attributes.
>   - **Message Queue**:
>     - Modifies `process_queue_spans()` and `process_spans_and_events_batch()` in `traces/consumer.rs` to handle batches of `RabbitMqSpanMessage`.
>     - Updates `push_spans_to_queue()` in `traces/producer.rs` to send batches of spans to RabbitMQ.
>   - **Span Handling**:
>     - Adds `project_id` to `Span` struct in `db/spans.rs`.
>     - Refactors `parse_and_enrich_attributes()` in `traces/spans.rs` for improved attribute handling.
>   - **Miscellaneous**:
>     - Removes `project_id` from `RabbitMqSpanMessage` in `api/v1/traces.rs`.
>     - Adjusts `create_span()` in `routes/spans.rs` to accommodate batch processing.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=lmnr-ai%2Flmnr&utm_source=github&utm_medium=referral)<sup> for 1d6d35a5bfbfd26bc294d903186ddb4309717ad6. You can [customize](https://app.ellipsis.dev/lmnr-ai/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/app-server/src/traces/consumer.rs
+++ b/app-server/src/traces/consumer.rs
@@ -385,41 +385,42 @@ async fn process_batch(
         }
     }
 
-    for span in stripped_spans {
-        // Push to evaluators queue - get evaluators for this span
-        match get_evaluators_by_path(&db, span.project_id, span.path).await {
-            Ok(evaluators) => {
-                if !evaluators.is_empty() {
-                    let span_output = span.output.clone().unwrap_or(Value::Null);
+    // FIXME: Temporarily commenting this out, while online evaluators cache is not implemented
+    // for span in stripped_spans {
+    //     // Push to evaluators queue - get evaluators for this span
+    //     match get_evaluators_by_path(&db, span.project_id, span.path).await {
+    //         Ok(evaluators) => {
+    //             if !evaluators.is_empty() {
+    //                 let span_output = span.output.clone().unwrap_or(Value::Null);
 
-                    for evaluator in evaluators {
-                        if let Err(e) = push_to_evaluators_queue(
-                            span.span_id,
-                            span.project_id,
-                            evaluator.id,
-                            span_output.clone(),
-                            evaluators_queue.clone(),
-                        )
-                        .await
-                        {
-                            log::error!(
-                                "Failed to push to evaluators queue. span_id [{}], project_id [{}]: {:?}",
-                                span.span_id,
-                                span.project_id,
-                                e
-                            );
-                        }
-                    }
-                }
-            }
-            Err(e) => {
-                log::error!(
-                    "Failed to get evaluators by path. span_id [{}], project_id [{}]: {:?}",
-                    span.span_id,
-                    span.project_id,
-                    e
-                );
-            }
-        }
-    }
+    //                 for evaluator in evaluators {
+    //                     if let Err(e) = push_to_evaluators_queue(
+    //                         span.span_id,
+    //                         span.project_id,
+    //                         evaluator.id,
+    //                         span_output.clone(),
+    //                         evaluators_queue.clone(),
+    //                     )
+    //                     .await
+    //                     {
+    //                         log::error!(
+    //                             "Failed to push to evaluators queue. span_id [{}], project_id [{}]: {:?}",
+    //                             span.span_id,
+    //                             span.project_id,
+    //                             e
+    //                         );
+    //                     }
+    //                 }
+    //             }
+    //         }
+    //         Err(e) => {
+    //             log::error!(
+    //                 "Failed to get evaluators by path. span_id [{}], project_id [{}]: {:?}",
+    //                 span.span_id,
+    //                 span.project_id,
+    //                 e
+    //             );
+    //         }
+    //     }
+    // }
 }


app-server/src/ch/spans.rs
```diff
@@ -116,22 +120,34 @@ impl CHSpan {
     }
 }
 
-pub async fn insert_span(clickhouse: clickhouse::Client, span: &CHSpan) -> Result<()> {
+pub async fn insert_spans_batch(clickhouse: clickhouse::Client, spans: &[CHSpan]) -> Result<()> {
+    if spans.is_empty() {
+        return Ok(());
+    }
+
     let ch_insert = clickhouse.insert("spans");
```
skull8888888, line 128: .with_option("async_insert", "1"); ?

dinmukhamedm: @skull8888888 This is already set globally in [main.rs:319](https://github.com/lmnr-ai/lmnr/blob/dev/app-server/src/main.rs#L319-L320), but we can be more explicit here in case we ever change the global setting.

app-server/src/db/spans.rs
```diff
@@ -100,25 +98,31 @@ pub async fn record_span(pool: &PgPool, span: &Span, project_id: &Uuid) -> Resul
...
+            .push_bind(values.input_preview)
+            .push_bind(values.output_preview)
+            .push_bind(&span.input_url as &Option<String>)
+            .push_bind(&span.output_url as &Option<String>)
+            .push_bind(&span.status)
+            .push_bind(&span.project_id);
+    });
+
+    query_builder.push(
+        " ON CONFLICT (span_id, project_id) DO UPDATE SET
```
skull8888888: why do we do on conflict here?

app-server/src/traces/consumer.rs
```diff
@@ -86,85 +102,329 @@ async fn inner_process_queue_spans(
...
+        let mut project_spans: HashMap<Uuid, Vec<RabbitMqSpanMessage>> = HashMap::new();
+        for message in rabbitmq_span_messages {
+            project_spans
+                .entry(message.span.project_id)
+                .or_insert_with(Vec::new)
+                .push(message);
+        }
+
+        // Check usage limits for each project
         if is_feature_enabled(Feature::UsageLimit) {
```
skull8888888: why do we check usage limits on consumer? I think it makes sense to do it in the producer

app-server/src/db/spans.rs
dinmukhamedm: I kept it the way it has been before. I think this is for cases if something's gone wrong and we accidentally tried to write the same span again. We assume the latest occurrence is the source of truth

skull8888888: isn't it less performant?

app-server/src/traces/consumer.rs
```diff
@@ -86,85 +102,329 @@ async fn inner_process_queue_spans(
...
+            match serde_json::from_slice::<Vec<RabbitMqSpanMessage>>(&delivery.data()) {
+                Ok(messages) => messages,
                 Err(e) => {
                     log::error!("Failed to deserialize span message: {:?}", e);
                     let _ = acker.reject(false).await;
                     continue;
                 }
             };
 
+        // Group spans by project_id to handle usage limits efficiently
```
skull8888888: we can avoid this grouping all together if we do in the producer for loop for each span.

app-server/src/traces/consumer.rs
```diff
@@ -86,85 +102,329 @@ async fn inner_process_queue_spans(
...
+
+        span_usages.push(span_usage);
+        trace_attributes.push(trace_attrs);
+    }
+
+    // Step 1: Record spans and traces to database (batch write)
+    match record_spans_batch(db.clone(), &spans, trace_attributes).await {
+        Ok(_) => {
+            // Step 2: Acknowledge the batch immediately after successful database write
+            let _ = acker.ack().await.map_err(|e| {
```
skull8888888: maybe we should ack after clickhouse insert? because spans are still in memory and another ones are already popped from the queue. Maybe that was the reason for unbounded memory rise?

app-server/src/db/spans.rs
dinmukhamedm: Depending on what to compare to.
- If we want to change this to `ON CONFLICT DO NOTHING`, i.e treat the first occurence as the source of truth, then yes, of course it's less performant. 
- If we want to retry or do something here in code in case of conflicts (why?), then postgres approach is quicker

I think let's do `on conflict do nothing`. What do you think?

skull8888888: let's just remove it completely? in what cases we can double write the span?

PR #717, opened 2025-07-17
batch span db write

skull8888888:
> [!IMPORTANT]
> Refactor span recording to batch writes in `record_spans()` with parallel processing and improved error handling.
> 
>   - **Behavior**:
>     - Refactor span recording to batch writes in `record_spans()` in `utils.rs`.
>     - Use `DB_WRITE_SPAN_BATCH_SIZE` environment variable to determine batch size.
>     - Process batches in parallel using `join_all`.
>   - **Functions**:
>     - Rename `record_spans_batch()` to `record_spans()` in `consumer.rs`.
>     - Add `record_spans()` in `utils.rs` to handle batching and parallel processing.
>   - **Error Handling**:
>     - Log errors for each failed batch in `record_spans()` in `utils.rs`.
>     - Return error if any batch fails in `record_spans()` in `utils.rs`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=lmnr-ai%2Flmnr&utm_source=github&utm_medium=referral)<sup> for 8ba26a59e8a941cec44a91b65cf8ef01e4dc9d31. You can [customize](https://app.ellipsis.dev/lmnr-ai/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/app-server/src/traces/utils.rs
+++ b/app-server/src/traces/utils.rs
@@ -1,6 +1,7 @@
 use std::sync::Arc;
 
 use backoff::ExponentialBackoffBuilder;
+use futures_util::future::join_all;
 use indexmap::IndexMap;
 use regex::Regex;
 use serde_json::Value;
@@ -22,6 +23,8 @@ use super::{
     attributes::TraceAttributes,
     spans::{SpanAttributes, SpanUsage},
 };
+
+const BATCH_SIZE: usize = 25;
 
 /// Calculate usage for both default and LLM spans
 pub async fn get_llm_usage_for_span(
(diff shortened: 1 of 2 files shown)


app-server/src/traces/utils.rs
```diff
@@ -23,6 +24,8 @@ use super::{
     spans::{SpanAttributes, SpanUsage},
 };
 
+const BATCH_SIZE: usize = 25;
```
dinmukhamedm: should we control this from env and default to 25 if not set?

app-server/src/traces/utils.rs
```diff
@@ -74,6 +77,39 @@ pub async fn get_llm_usage_for_span(
...
+            log::error!("Failed to record spans: {:?}", e);
+            errors.push(e);
+        }
+    }
+
+    if !errors.is_empty() {
+        return Err(anyhow::anyhow!("Failed to record some spans: {:?}", errors));
+    }
+
+    Ok(())
```
dinmukhamedm: nit: this is extremely non-idiomatic Rust, we could iter().map(), and I think we could even iterate over `Vec<Result>` and collect it as `<Result<Vec<T>, Error>>` or even `<Result<Vec<T>, Vec<E>>>` with just the right type hints. No need to go all the way Rust, but at least let's remove mut and for

PR #719, opened 2025-07-17
feat: spans API LAM-745

olzhik11:
- /spans search endpoint
- /spans/:span_id endpoint


app-server/src/db/span.rs
```diff
@@ -0,0 +1,127 @@
...
+            s.parent_span_id, 
+            s.name,
+            s.attributes,
+            s.input,
+            s.output,
+            s.span_type,
+            s.start_time, 
+            s.end_time, 
+            s.status,
+            s.input_url,
```
olzhik11, line 109: remove input output url

PR #754, opened 2025-07-25
Add columns to ch spans

skull8888888:


app-server/src/ch/spans.rs
```diff
@@ -101,6 +107,14 @@ impl CHSpan {
                 .clone()
                 .or(usage.request_model.clone())
                 .unwrap_or(String::from("<null>")),
+            request_model: usage
+                .request_model
+                .clone()
+                .unwrap_or(String::from("<null>")),
```
dinmukhamedm: shall we use empty strings maybe? No strong opinion though

app-server/src/db/spans.rs
```diff
@@ -68,54 +65,33 @@ pub struct Span {
...
+    let input_previews: Vec<Option<String>> = spans
+        .par_iter()
+        .map(|s| get_string_preview(&s.input))
         .collect();
-    let inputs: Vec<Option<Value>> = span_values
-        .iter()
-        .map(|v| v.sanitized_input.clone())
+    let output_previews: Vec<Option<String>> = spans
+        .par_iter()
+        .map(|s| get_string_preview(&s.output))
```
dinmukhamedm: 1. IMPORTANT: still need to sanitize here (or in the `get_string_preview`)
2. Didn't you mention you were going to remove this altogether and query from clickhouse on the frontend? Or are we doing this migration step by step?

app-server/src/db/utils.rs
```diff
@@ -12,37 +14,15 @@ pub fn span_id_to_uuid(span_id: &[u8]) -> Uuid {
...
-                return false;
-            }
-            // Remove other control characters
-            if c.is_control() {
-                return false;
-            }
-            true
-        })
-        .collect::<String>()
-}
```
dinmukhamedm, line 35: curious if clickhouse requires anything similar. Can you try sending data with such character to clickhouse and actually wait for async insert, see what happens?

app-server/src/db/spans.rs
skull8888888: 1. Do we need to sanitize string? I thought it was because of jsonb.
2. I wanted, but we need to migrate spans tab first

PR #759, opened 2025-07-28
join spans and events in memory

dinmukhamedm:
> [!IMPORTANT]
> Refactor `getSharedSpans` and `getTraceSpans` to join spans and events in memory for improved performance.
> 
>   - **Behavior**:
>     - Refactor `getSharedSpans` in `shared/spans/index.ts` to join spans and events in memory using `groupBy`.
>     - Refactor `getTraceSpans` in `spans/index.ts` to join spans and events in memory using `groupBy`.
>     - Remove SQL-based JSON aggregation and joins for spans and events.
>   - **Performance**:
>     - Improve performance by avoiding slow PostgreSQL JSON aggregation and joins.
>     - Ensure correct index usage by moving join logic to in-memory processing.
>   - **Imports**:
>     - Add `groupBy` from `lodash` in both `shared/spans/index.ts` and `spans/index.ts`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=lmnr-ai%2Flmnr&utm_source=github&utm_medium=referral)<sup> for 6c0d89ae232ff7be0216681a46fb976456b466ce. You can [customize](https://app.ellipsis.dev/lmnr-ai/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/frontend/lib/actions/spans/index.ts
+++ b/frontend/lib/actions/spans/index.ts
@@ -153,6 +153,10 @@ export async function getTraceSpans(input: z.infer<typeof GetTraceSpansSchema>) 
     )
     .orderBy(asc(spans.startTime));
 
+  if (spanItems.length === 0) {
+    return [];
+  }
+
   // Join in memory, because json aggregation and join in PostgreSQL may be too slow
   // depending on the number of spans and events, and there is no way for us
   // to force PostgreSQL to use the correct indexes always.
--- a/frontend/lib/actions/spans/index.ts
+++ b/frontend/lib/actions/spans/index.ts
@@ -1,5 +1,5 @@
 import { and, asc, desc, eq, getTableColumns, inArray, sql } from "drizzle-orm";
-import { compact, isNil } from "lodash";
+import { compact, groupBy, isNil } from "lodash";
 import { z } from "zod/v4";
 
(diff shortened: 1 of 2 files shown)


frontend/lib/actions/spans/index.ts
```diff
@@ -129,39 +129,7 @@ export async function getTraceSpans(input: z.infer<typeof GetTraceSpansSchema>)
...
-          )`
-          )
-        )
-      )
-      .groupBy(events.spanId, events.projectId)
-  );
-
   const spanItems = await db
-    .with(spanEventsQuery)
     .select({
```
skull8888888, line 133: interesting, now nothing stops us from querying spans from ch

frontend/lib/actions/spans/index.ts
```diff
@@ -173,13 +141,8 @@ export async function getTraceSpans(input: z.infer<typeof GetTraceSpansSchema>)
       attributes: spans.attributes,
```
skull8888888, line 141: Do we need return attributes?

dinmukhamedm: There is some logic that stores the span path in the trace view so that we jump to the same span if we `j`-`k` between identical traces. Could optimize away, but I think it's out of scope of this PR

PR #763, opened 2025-07-30
feat: custom dashboards LAM-761 (2)

olzhik11:


app-server/src/api/v1/sql.rs
```diff
@@ -2,13 +2,15 @@ use std::sync::Arc;
...
 use serde::Deserialize;
+use serde_json::Value;
 
 use crate::{db::project_api_keys::ProjectApiKey, routes::types::ResponseResult, sql};
 
 #[derive(Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct SqlQueryRequest {
     pub query: String,
+    pub parameters: Option<Value>,
```
dinmukhamedm: shall we enforce `HashMap<String, Value>`? In the query engine, we expect `dict`

frontend/lib/time.ts
```diff
@@ -0,0 +1,88 @@
...
+  })
+  .refine(
+    (data) =>
+      // If using absolute dates, both must be provided
+      !((data.startTime && !data.endTime) || (!data.startTime && data.endTime)),
+    {
+      message: "Both startDate and endDate must be provided when using absolute dates",
+      path: ["startDate", "endDate"],
+    }
+  )
```
dinmukhamedm: this is extremely hard to reason about. Is there no way to do that using a `z.enum` or something similar to serde untagged enum, so that we have TimeInputSchemaRelative and TimeInputSchemaAbsolute?

PR #788, opened 2025-08-11
feat: move workspace create, get to next

olzhik11:
> [!IMPORTANT]
> Moved workspace creation and retrieval from backend to frontend using Next.js and Drizzle ORM, with enhanced error handling via Zod.
> 
>   - **Backend Removals**:
>     - Removed `workspace` module from `db/mod.rs` and `routes/mod.rs`.
>     - Deleted `workspace.rs` from `db` and `routes`.
>     - Removed `validator` function from `auth/mod.rs`.
>     - Removed workspace-related routes from `main.rs`.
>   - **Frontend Additions**:
>     - Added `createWorkspace` and `getWorkspaces` functions in `lib/actions/workspaces/index.ts` using Drizzle ORM.
>     - Implemented `GET` and `POST` handlers in `api/workspaces/route.ts` for workspace operations.
>     - Added `GET` handler in `api/workspaces/[workspaceId]/route.ts` for fetching a specific workspace.
>     - Utilized Zod for input validation and error handling in `route.ts` files.
>   - **Miscellaneous**:
>     - Updated `WorkspaceTier` enum in `lib/workspaces/types.ts` to include `HOBBY` tier.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=lmnr-ai%2Flmnr&utm_source=github&utm_medium=referral)<sup> for adf6f12109d2ac1b0f233a81c9cad295607eca10. You can [customize](https://app.ellipsis.dev/lmnr-ai/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


app-server/src/auth/mod.rs
```diff
@@ -39,37 +39,6 @@ impl FromRequest for ProjectApiKey {
     }
 }
 
-pub async fn validator(
```
skull8888888, line 42: are we 100% sure we're safe to remove it? That means we no longer do any crud in app-server? What about project api key generation?

frontend/lib/actions/workspaces/index.ts
```diff
@@ -1,13 +1,212 @@
...
+          name: projects.name,
+          workspaceId: projects.workspaceId,
+        })
+        .from(projects)
+        .where(eq(projects.workspaceId, workspace.id));
+
+      return {
+        id: workspace.id,
+        name: workspace.name,
+        tierName: workspace.tierName as "Free" | "Pro",
```
dinmukhamedm: "Hobby"? Let's define this type elsewhere and use the type here.

frontend/lib/actions/workspaces/index.ts
```diff
@@ -1,13 +1,212 @@
...
+    })
+    .from(users)
+    .innerJoin(membersOfWorkspaces, eq(users.id, membersOfWorkspaces.userId))
+    .where(eq(membersOfWorkspaces.workspaceId, workspaceId));
+
+  const workspaceUserInfos: WorkspaceUser[] = workspaceUsers.map((user) => ({
+    id: user.id,
+    name: user.name ?? "",
+    email: user.email ?? "",
+    role: user.role ?? "member",
```
dinmukhamedm: is this a drizzle issue? Why and how could these fields be `undefined`?

frontend/lib/actions/workspaces/index.ts
```diff
@@ -1,13 +1,212 @@
...
+
+  return {
+    id: workspace[0].id,
+    name: workspace[0].name,
+    tierName: workspace[0].tierName as WorkspaceTier,
+    users: workspaceUserInfos,
+  };
+};
+
+export const addOwnerToWorkspace = async (userId: string, workspaceId: string) => {
```
dinmukhamedm: is this used anywhere? I think you do that above in line 74, and it's better to do it inline (because there is no other scenario where we would add an owner to a workspace).

app-server/src/auth/mod.rs
dinmukhamedm: This was already not behind the validator, because `/api` is not exposed, and `/api/v1/projects` is (supposedly) protected by Next.js middleware. I think this auth isn't used anywhere any longer

olzhik11: for project api key, auth on path projects/{project_id} is handled by middleware on Next.js

frontend/lib/actions/workspaces/index.ts
olzhik11: makes sense, we don't need coalescing op. here

frontend/lib/actions/workspaces/index.ts
olzhik11: good catch

frontend/lib/actions/workspaces/index.ts
```diff
@@ -1,13 +1,196 @@
...
+    .select({
+      id: users.id,
+      name: users.name,
+      email: users.email,
+      role: membersOfWorkspaces.memberRole,
+      createdAt: membersOfWorkspaces.createdAt,
+    })
+    .from(users)
+    .innerJoin(membersOfWorkspaces, eq(users.id, membersOfWorkspaces.userId))
+    .where(eq(membersOfWorkspaces.workspaceId, workspaceId))) as WorkspaceUser[];
```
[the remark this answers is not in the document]
olzhik11: they are not null

dinmukhamedm (approved):

PR #793, opened 2025-08-12
fix limits in browser events

dinmukhamedm:
> [!IMPORTANT]
> Add feature flag check for usage limits in `create_session_event` to return 403 if exceeded.
> 
>   - **Behavior**:
>     - Add feature flag check for `Feature::UsageLimit` in `create_session_event` in `browser_sessions.rs`.
>     - If `UsageLimit` is enabled and workspace data limit is exceeded, return 403 Forbidden response.
>   - **Dependencies**:
>     - Add `db`, `cache`, and `clickhouse` as parameters to `create_session_event` to support limit checking.
>   - **Imports**:
>     - Import `DB`, `is_feature_enabled`, and `get_workspace_limit_exceeded_by_project_id` in `browser_sessions.rs`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=lmnr-ai%2Flmnr&utm_source=github&utm_medium=referral)<sup> for b52976fda6a8bb517425fbbe5c49163a37be7974. You can [customize](https://app.ellipsis.dev/lmnr-ai/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/app-server/src/api/v1/browser_sessions.rs
+++ b/app-server/src/api/v1/browser_sessions.rs
@@ -96,7 +96,7 @@ async fn create_session_event(
         .await?;
 
         if limits_exceeded.bytes_ingested {
-            return Ok(HttpResponse::PaymentRequired().json("Workspace data limit exceeded"));
+            return Ok(HttpResponse::Forbidden().json("Workspace data limit exceeded"));
         }
     }
 
--- a/app-server/src/api/v1/traces.rs
+++ b/app-server/src/api/v1/traces.rs
@@ -47,7 +47,7 @@ pub async fn process_traces(
         .await?;
 
         if limits_exceeded.bytes_ingested {
-            return Ok(HttpResponse::PaymentRequired().json("Workspace data limit exceeded"));
+            return Ok(HttpResponse::Forbidden().json("Workspace data limit exceeded"));
         }
     }
 


app-server/src/api/v1/browser_sessions.rs
```diff
@@ -81,6 +86,20 @@ async fn create_session_event(
...
+        let limits_exceeded = get_workspace_limit_exceeded_by_project_id(
+            db.into_inner(),
+            clickhouse.into_inner().as_ref().clone(),
+            cache.into_inner(),
+            project_api_key.project_id,
+        )
+        .await?;
+
+        if limits_exceeded.bytes_ingested {
+            return Ok(HttpResponse::PaymentRequired().json("Workspace data limit exceeded"));
```
skull8888888: are you sure payment required is a better response than forbidden?

dinmukhamedm: not much preference. 

403 Forbidden is generally used when operation is not allowed because not enough permissions. 

402 Payment Required is not dedicated for anything and was initially created for payment systems, but there is no convention. I read somewhere that GCP uses this code when you need to upgrade.

If you feel like the more conventional, better documented API response is better, I can update, no problem

skull8888888: I think 403 is better for now? I've never seen 402 in the wild.

dinmukhamedm: Ok, changed to 403 Forbidden

PR #831, opened 2025-09-05
fix: replace postgres with clickhouse operations

olzhik11:
> [!IMPORTANT]
> Replace PostgreSQL operations with ClickHouse operations for span handling and refactor utility functions across multiple files.
> 
>   - **Database Operations**:
>     - Replace PostgreSQL operations with ClickHouse operations in `updateTraceVisibility()` in `trace/index.ts`.
>     - Use ClickHouse for querying spans in `getSharedSpan()` in `shared/span/index.ts` and `getSpan()` in `span/index.ts`.
>   - **Utility Functions**:
>     - Move `tryParseJson()` to `common/utils.ts` and use it in `shared/span/index.ts`, `span/index.ts`, and `trace/index.ts`.
>   - **Misc**:
>     - Remove unused imports and code related to PostgreSQL spans in `trace/index.ts`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=lmnr-ai%2Flmnr&utm_source=github&utm_medium=referral)<sup> for ea2e06563f0c639736bb0180e3bd731368b66f33. You can [customize](https://app.ellipsis.dev/lmnr-ai/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/frontend/lib/actions/common/utils.ts
+++ b/frontend/lib/actions/common/utils.ts
@@ -19,9 +19,7 @@ export const processFilters = <TFilter extends DatatableFilter = DatatableFilter
 
   return filters.flatMap((filter) => {
     const processor =
-      processors.get(`${filter.column}:${filter.operator}`) ||
-      processors.get(filter.column) ||
-      defaultProcessor;
+      processors.get(`${filter.column}:${filter.operator}`) || processors.get(filter.column) || defaultProcessor;
 
     if (!processor) return [];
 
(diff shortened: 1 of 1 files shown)


frontend/lib/actions/common/utils.ts
```diff
@@ -60,3 +60,19 @@ export const parseUrlParams = <T>(
 
   return schema.safeParse(obj);
 };
+
+export const tryParseJson = (value: string) => {
```
dinmukhamedm, line 66: nit: perhaps we should reflect that this function has this special behavior, especially now that it's inside common utils. The mere reason why we had to invent this function in the frontend is that in the backend we moved a function with special behavior to the common utils and then forgot about it, so let's avoid this

dinmukhamedm (dismissed):

PR #869, opened 2025-09-17
updates to clickhouse eval scores, tags

dinmukhamedm:


app-server/src/ch/spans.rs
```diff
@@ -186,3 +191,45 @@ pub async fn insert_spans_batch(clickhouse: clickhouse::Client, spans: &[CHSpan]
...
+    }
+
+    let query_result = clickhouse
+        .query("SELECT tags FROM spans WHERE span_id = ? AND project_id = ?")
+        .bind(span_id)
+        .bind(project_id)
+        .fetch_one::<String>()
+        .await?;
+
+    let existing_tags = serde_json::from_str::<HashSet<String>>(&query_result)?;
```
olzhik11, line 212: just to confirm, this will resolve to `HashSet<String>` if stored as stringified `Vec<String>`, right?

olzhik11 (approved):
besides existing comments looks fine

app-server/src/ch/spans.rs
dinmukhamedm: Yes. https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=350decbd0c6e46c5cf0ccfe9fc9a41b1

PR #910, opened 2025-09-26
only query spans table to check

skull8888888:
> [!IMPORTANT]
> Update SQL query in `page.tsx` to select from `spans` table using `project_id` and `trace_type = 0`.
> 
>   - **Query Update**:
>     - In `page.tsx`, change SQL query to select from `spans` table instead of `traces`.
>     - Modify query condition to use `project_id` and `trace_type = 0`.
>     - Remove `traceType` parameter and use `projectId` instead.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=lmnr-ai%2Flmnr&utm_source=github&utm_medium=referral)<sup> for 7bbb3257f4337f8500cabab197e4258e10f75ebb. You can [customize](https://app.ellipsis.dev/lmnr-ai/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/frontend/app/project/[projectId]/traces/page.tsx
+++ b/frontend/app/project/[projectId]/traces/page.tsx
@@ -18,12 +18,13 @@ export default async function TracesPage(props: { params: Promise<{ projectId: s
   const [result] = await executeQuery<{ exists: number }>({
     query: `
         SELECT 1 as exists
-        FROM traces
-        WHERE trace_type = {traceType:String}
+        FROM spans
+        WHERE project_id = {projectId:UUID}
+          AND trace_type = 0
         LIMIT 1
     `,
     parameters: {
-      traceType: "DEFAULT",
+      projectId,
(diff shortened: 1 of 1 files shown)


frontend/app/project/[projectId]/traces/page.tsx
```diff
@@ -18,12 +18,13 @@ export default async function TracesPage(props: { params: Promise<{ projectId: s
   const [result] = await executeQuery<{ exists: number }>({
     query: `
         SELECT 1 as exists
-        FROM traces
-        WHERE trace_type = {traceType:String}
+        FROM spans
```
Comment (dinmukhamedm) [inline frontend/app/project/[projectId]/traces/page.tsx:21]: there's a similar query in the evals page, let's update it too. In there, we can probably not check for trace_type

dinmukhamedm (approved):

PR #914, opened 2025-09-28
Fix/traces

skull8888888:


app-server/src/ch/traces.rs
```diff
@@ -0,0 +1,220 @@
...
+
+impl CHTrace {
+    /// Create CHTrace from database Trace with additional fields
+    pub fn from_db_trace(
+        trace: &Trace,
+        user_id: String,
+        top_span_id: Uuid,
+        top_span_name: String,
+        top_span_type: u8,
+        trace_type: u8,
```
dinmukhamedm: nit: I think we have stronger type enums for this in src/ch/spans.rs, let's re-use that

app-server/src/traces/consumer.rs
```diff
@@ -302,6 +302,46 @@ async fn process_batch(
...
+    // Process trace aggregations and update trace statistics
+    let trace_aggregations = TraceAggregation::from_ch_spans(&ch_spans);
+    if !trace_aggregations.is_empty() {
+        // Upsert trace statistics in PostgreSQL
+        match upsert_trace_statistics_batch(&db.pool, &trace_aggregations).await {
+            Ok(updated_traces) => {
+                // Convert to ClickHouse traces and upsert
+                let ch_traces: Vec<CHTrace> = updated_traces
+                    .iter()
+                    .zip(trace_aggregations.iter())
```
dinmukhamedm: not sure, but this looks like it would break if there are spans from more than one trace in the batch? I think the `from_ch_spans` could return the full `HashMap` instead of `into_values`, and we should look up here.

app-server/src/db/trace.rs
```diff
@@ -58,3 +111,92 @@ pub async fn update_trace_type(
...
+                total_token_count, 
+                input_cost, 
+                output_cost, 
+                cost,
+                status,
+                tags,
+                num_spans
+            )
+            VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14, $15)
+            ON CONFLICT (id) DO UPDATE SET
```
dinmukhamedm: I think we should recreate the primary key on table traces now to be project_id, id. If we do so, we'll need to update the `ON CONFLICT` clause.

Might also be a good time to make start_time and end_time non-nullable in the table?

PR #1041, opened 2025-11-19
wip: search spans with quickwit

Rainhunter13:
> [!NOTE]
> Integrates Quickwit to index spans and power text search, adding a spans indexer pipeline, a search API, frontend search wiring, and docker/env setup.
> 
> - **Backend**:
>   - **Quickwit integration**: add `quickwit/` (gRPC ingest + HTTP search client, proto, doc batch, producer/consumer) and compile proto in `build.rs`.
>   - **Indexing pipeline**: publish spans to `SPANS_INDEXER_*` MQ; new consumer `process_indexer_queue_spans` ingests to Quickwit; add optional `QuickwitClient` from `QUICKWIT_*` env.
>   - **API**: add `POST /api/v1/projects/{project_id}/spans/search` using Quickwit; inject client into app state.
>   - **Workers/health**: extend `WorkerType` with `SpansIndexer`; track expected counts.
>   - **Config**: add `QUICKWIT_INGEST_URL`, `QUICKWIT_SEARCH_URL`, `QUICKWIT_SPANS_INDEX_ID` to `.env` and wiring in `main.rs`.
> - **Frontend**:
>   - **Search plumbing**: new `lib/actions/traces/search.ts` calling backend search; refactor traces/spans/evaluation queries to use Quickwit hits; disable infinite scrolling during text search; use `DEFAULT_SEARCH_MAX_HITS`.
>   - **Initialization**: initialize Quickwit indexes at startup; add `lib/quickwit/indexes/spans.yaml`.
>   - **Env**: expose `QUICKWIT_SEARCH_URL` in `.env.local.example`.
> - **Ops**:
>   - **Docker**: add `quickwit` service with ports/volume; propagate `QUICKWIT_*` env across compose files.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 58d728aae163183f4df61f255128c0ade073440e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


app-server/src/traces/consumer.rs
```diff
@@ -502,6 +563,222 @@ async fn process_batch(
...
+        ));
+    }
+
+    queue
+        .publish(&payload, SPANS_INDEXER_EXCHANGE, SPANS_INDEXER_ROUTING_KEY)
+        .await
+        .context("Failed to publish spans to Quickwit indexer queue")?;
+
+    Ok(())
+}
```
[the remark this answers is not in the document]
Rainhunter13: => should split into smaller chunks in case payload is too big

But, if any other error during publishing, should we
- ack anyway and allow potential missing entries in the search engine
- reject the whole batch (not ack), but then either clickhouse writes will be rejected as well, or, if written first, will be duplicated upon re-reading those messages
- do some different queue logic separating clickhouse and quickwit

🤔

app-server/src/traces/consumer.rs
```diff
@@ -243,6 +254,45 @@ struct StrippedSpan {
...
+#[derive(Clone)]
+pub struct QuickwitIngestConfig {
+    pub endpoint: String,
+}
+
+impl QuickwitIngestConfig {
+    pub fn from_env() -> Self {
+        Self {
+            endpoint: env::var("QUICKWIT_INGEST_URL")
+                .unwrap_or_else(|_| "http://localhost:7281".to_string()),
```
dinmukhamedm: nit: it's fine for you not to know this, but not fine for coding agents. Yet, none of them realize that:
```
.unwrap_or_else(|_| T)
```

is identical to

```
.unwrap_or(T)
```

That's something we'll have to fight for for another few years Ig

app-server/src/traces/consumer.rs
```diff
@@ -386,6 +436,17 @@ async fn process_batch(
     // Send realtime span updates directly to SSE connections after successful ClickHouse writes
     send_span_updates(&spans, &sse_connections).await;
 
+    let quickwit_spans: Vec<QuickwitIndexedSpan> =
+        spans.iter().map(QuickwitIndexedSpan::from).collect();
```
dinmukhamedm: nit: I think Into and from are interchangeable, so I'd prefer

```
let quickwit_spans: Vec<QuickwitIndexedSpan> =
    spans.iter().map(|span| span.into()).collect();
```

app-server/src/traces/consumer.rs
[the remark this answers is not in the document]
dinmukhamedm: > ack anyway and allow potential missing entries in the search engine

app-server/src/traces/consumer.rs
```diff
@@ -502,6 +563,222 @@ async fn process_batch(
...
+        {
+            log::error!(
+                "Quickwit spans indexer worker exited with error: {:?}. Retrying after backoff...",
+                e
+            );
+        } else {
+            log::warn!("Quickwit spans indexer worker exited gracefully. Restarting...");
+        }
+
+        sleep(Duration::from_secs(5)).await;
```
dinmukhamedm: do we have this in the other queues?

app-server/src/traces/consumer.rs
```diff
@@ -502,6 +563,222 @@ async fn process_batch(
...
+                "Failed to connect to Quickwit spans indexer queue after retries: {:?}",
+                e
+            );
+            return Err(anyhow::anyhow!(
+                "Could not bind to Quickwit spans indexer queue"
+            ));
+        }
+    };
+
+    let mut quickwit_client = connect_quickwit_client(&quickwit_config.endpoint).await?;
```
dinmukhamedm: I think client must be configured in main.rs and passed here in an Arc

dinmukhamedm: Also, why mut and reassign in line 731?

app-server/src/traces/consumer.rs
```diff
@@ -502,6 +563,222 @@ async fn process_batch(
...
+            }
+        }
+    }
+
+    log::warn!("Quickwit spans indexer queue closed connection");
+    Ok(())
+}
+
+async fn ingest_spans_into_quickwit(
+    client: &mut IngestServiceClient<Channel>,
```
dinmukhamedm: does it have to be mut to call .ingest?

app-server/src/traces/quickwit_doc_batch.rs
```diff
@@ -0,0 +1,99 @@
+// Portions of this module are adapted from Quickwit's Apache-2.0 licensed
+// implementation in quickwit-ingest/src/doc_batch.rs.
```
dinmukhamedm: nit: let's properly link github?

frontend/instrumentation.ts
```diff
@@ -77,6 +77,19 @@ export async function register() {
...
+        try {
+          const { initializeQuickwitIndexes } = await import("@/lib/quickwit/migrations.ts");
+          await initializeQuickwitIndexes();
+        } catch (error) {
+          console.error("Failed to initialize Quickwit indexes:", error);
+          // Don't throw - Quickwit might not be available in all environments
+          console.log("Continuing without Quickwit indexes...");
+        }
+      };
+      await initializeQuickwit();
```
dinmukhamedm, line 95: probs need to check for the presence of QUICKWIT_CONTROL_PLANE_URL env var

app-server/src/traces/consumer.rs
dinmukhamedm, line 1: Like Robert mentioned offline, let's move all of the quickwit related stuff from here into a separate module, e.g. src/quickwit

app-server/src/traces/consumer.rs
Rainhunter13: nope, removed

app-server/src/traces/consumer.rs
Rainhunter13: yes because `ingest` expects mutable `self`
```
pub async fn ingest(
            &mut self,
            request: impl tonic::IntoRequest<super::IngestRequest>,
        )
```

frontend/instrumentation.ts
Rainhunter13: do you mean so that we can print error and not init quickwit rather than using the default one defined under `quickwit/migrations.ts`?

dinmukhamedm: yes

app-server/src/quickwit/client.rs
```diff
@@ -0,0 +1,128 @@
...
+    pub span_id: Uuid,
+    pub project_id: Uuid,
+    pub trace_id: Uuid,
+    pub start_time: DateTime<Utc>,
+    pub input: Option<String>,
+    pub output: Option<String>,
+    pub attributes: Value,
+}
+
+impl From<&Span> for QuickwitIndexedSpan {
```
dinmukhamedm: This should probably go in quickwit/mod.rs?

app-server/src/main.rs
```diff
@@ -590,6 +625,11 @@ fn main() -> anyhow::Result<()> {
             .parse::<u8>()
             .unwrap_or(4);
 
+        let num_spans_indexer_workers = env::var("NUM_SPANS_INDEXER_WORKERS")
+            .unwrap_or(String::from("1"))
+            .parse::<u8>()
+            .unwrap_or(1);
```
dinmukhamedm: nit: let's do 4

docker-compose-local-dev-full.yml
```diff
@@ -64,7 +64,20 @@ services:
       timeout: 5s
       retries: 5
 
+  quickwit:
```
dinmukhamedm, line 67: let's add this to all docker compose files

app-server/src/traces/consumer.rs
```diff
@@ -502,6 +516,176 @@ async fn process_batch(
     }
 }
 
+async fn publish_spans_for_indexing(
```
dinmukhamedm: let's move this to quickwit/producer.rs

app-server/src/traces/consumer.rs
```diff
@@ -502,6 +516,176 @@ async fn process_batch(
...
+
+    queue
+        .publish(&payload, SPANS_INDEXER_EXCHANGE, SPANS_INDEXER_ROUTING_KEY)
+        .await
+        .context("Failed to publish spans to Quickwit indexer queue")?;
+
+    Ok(())
+}
+
+pub async fn process_queue_spans_indexer(
```
dinmukhamedm: and this to quickwit/consumer.rs. Also, how about we rename this to something like `process_indexer_queue_span(s)`?

app-server/src/routes/spans.rs
```diff
@@ -92,3 +93,115 @@ pub async fn create_span(
...
+        search_body["start_timestamp"] = json!(start.timestamp());
+    }
+    if let Some(end) = request.end_time {
+        search_body["end_timestamp"] = json!(end.timestamp());
+    }
+
+    // Handle pagination
+    if request.limit != 0 {
+        search_body["max_hits"] = json!(request.limit);
+    }
```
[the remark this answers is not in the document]
Rainhunter13: nope

app-server/src/quickwit/consumer.rs
```diff
@@ -0,0 +1,185 @@
...
+            continue;
+        }
+
+        for span in &mut indexed_spans {
+            // Extract text content from JSON value for searchability
+            // This avoids double-encoding: we want plain text, not a JSON-encoded string
+            let attributes_text = extract_text_from_json_value(&span.attributes);
+            let attributes_text = attributes_text.replace('{', " { ").replace('}', " } ");
+            span.attributes = serde_json::Value::String(attributes_text);
+        }
```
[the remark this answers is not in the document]
Rainhunter13: wdym there are normal strings

app-server/src/quickwit/consumer.rs
```diff
@@ -0,0 +1,192 @@
...
+
+        match quickwit_client.ingest("spans", &indexed_spans).await {
```
dinmukhamedm: super nit: index name should probably at least be a const at the top of the file. Ideally an env var defaulting to a const at the top of the file

app-server/src/routes/spans.rs
```diff
@@ -92,3 +95,131 @@ pub async fn create_span(
...
+    let mut query_parts = vec![
+        format!("project_id:{}", project_id),
+        format!("({})", trimmed_query),
+    ];
+
+    let mut sort_by = "_score,start_time"; // default sort for scores and timestamp in quickwit is desc!
+
+    if !request.trace_id.is_empty() {
+        query_parts.push(format!("trace_id:{}", request.trace_id));
+        sort_by = "start_time"; // sort by timestamp (desc) inside a single trace
```
dinmukhamedm, line 158: shouldn't this be asc? Or it doesn't matter since we now only take ids

app-server/src/routes/spans.rs
```diff
@@ -92,3 +95,131 @@ pub async fn create_span(
...
+    }
+    if let Some(end) = request.end_time {
+        search_body["end_timestamp"] = json!(end.timestamp());
+    }
+
+    // Handle pagination
+    if request.limit != 0 {
+        search_body["max_hits"] = json!(request.limit);
+    } else {
+        search_body["max_hits"] = json!(DEFAULT_SEARCH_MAX_HITS);
```
dinmukhamedm: nit: let's prefer things like
```
serde_json::Value::Number(request.limit.into());
serde_json::Value::String(search_field_str);
```
instead of `json!` macro

frontend/.env.local.example
```diff
@@ -36,4 +36,5 @@ S3_TRACE_PAYLOADS_BUCKET=
 REDIS_URL=
 OPENAI_API_KEY=
 
-QUERY_ENGINE_URL=
\ No newline at end of file
+QUERY_ENGINE_URL=
+QUICKWIT_SEARCH_URL=
```
dinmukhamedm, line 40: I think this should be added to the app-server as well?

app-server/src/routes/spans.rs
```diff
@@ -92,3 +95,131 @@ pub async fn create_span(
...
+    };
+
+    let mut query_parts = vec![
+        format!("project_id:{}", project_id),
+        format!("({})", trimmed_query),
+    ];
+
+    let mut sort_by = "_score,start_time"; // default sort for scores and timestamp in quickwit is desc!
+
+    if !request.trace_id.is_empty() {
```
dinmukhamedm: similar to the frontend, we could instead pass Option<Uuid> for trace_id. The way this works:

1. client passes nothing (undefined) to the action function.
2. `JSON.stringify({key1: "value1", key2: undefined})` results in `{"key1": "value1"}` i.e. undefined values are omitted
3. In rust we need to annotate the Option field on a struct with `#[serde(default)]` so that when the field is absent, it is filled with `Option::<T>::default()`, which is obviously None
4. In this line, we then check for `request.trace_id.is_none()`

frontend/lib/actions/traces/search.ts
```diff
@@ -0,0 +1,61 @@
...
+
+export const searchSpans = async ({
+  projectId,
+  traceId,
+  searchQuery,
+  timeRange,
+  searchType,
+}: {
+  projectId: string;
+  traceId: string;
```
dinmukhamedm: let's change this to `traceId?: string` and pass undefined instead of empty string.

app-server/src/routes/spans.rs
Rainhunter13: hm I think that's the current api behavior, we return spans in timestamp desc order

dinmukhamedm: Probably sorted in the frontend (not the best idea but works).

PR #1069, opened 2025-12-04
Workers

skull8888888:
> [!NOTE]
> Introduce a unified worker framework and refactor all MQ consumers into handler-based workers, updating wiring, routing keys, and health probes.
> 
> - **Worker Framework**
>   - Add `worker` module: `MessageHandler`, `HandlerError` (permanent/transient), `QueueWorker`, `WorkerPool`, `QueueConfig`, `WorkerType`.
>   - Centralize MQ connect/receive/ack/reject with backoff and requeue control.
> - **Consumers → Handlers**
>   - Spans: `traces/consumer` → `SpanHandler` (batch processing; ClickHouse write errors now bubble as handler errors).
>   - Browser events: `BrowserEventHandler` with backoff and usage-limit updates.
>   - Evaluators: `EvaluatorHandler` (HTTP call, DB/ClickHouse writes, error propagation).
>   - Payloads: `PayloadHandler` (store with backoff).
>   - Notifications: `NotificationHandler` (Slack integration lookup and send).
>   - Quickwit indexer: `QuickwitIndexerHandler` (attribute text extraction, transient requeue on ingest failure).
>   - Trace summaries: `TraceSummaryHandler` (summary/event-identification calls with retry; notifications; optional clustering enqueue).
>   - Clustering: `ClusteringHandler` (cache lock, transient timeout requeue, HTTP call).
> - **App Wiring**
>   - Replace per-consumer loops with `WorkerPool::spawn` across queues in `main.rs`.
>   - Add/use routing key constants for `observations`, `spans_indexer`, `browser_sessions`, `evaluators`, `payloads`, `trace_summary`, `notifications`, `clustering`.
> - **Probes/Health**
>   - Simplify `/health` and `/ready` to RabbitMQ connectivity; remove `worker_tracking` module and consumer-specific health endpoint.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a0c2b601d23e209a85d1c96c98edaf549e8b8735. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

----

> [!IMPORTANT]
> Introduce a unified worker framework, migrate queue consumers to handlers, update main wiring via WorkerPool, and remove legacy worker tracking/health.
> 
>   - **Core worker framework**:
>     - Add `worker/` with `MessageHandler`, `HandlerError`, `QueueWorker`, `WorkerPool`, `QueueConfig`, `WorkerType` for unified MQ processing and reconnect/backoff.
>   - **Consumers migrated to handlers**:
>     - `traces/consumer` → `SpanHandler`
>     - `quickwit/consumer` → `QuickwitIndexerHandler`
>     - `browser_events` → `BrowserEventHandler`
>     - `evaluators` → `EvaluatorHandler`
>     - `storage` → `PayloadHandler`
>     - `traces/summary` → `TraceSummaryHandler`
>     - `traces/clustering` → `ClusteringHandler`
>     - `notifications` → `NotificationHandler`
>   - **Application wiring**:
>     - Use `WorkerPool::spawn(...)` in `main.rs` for all queues with explicit routing keys (`OBSERVATIONS_*`, `SPANS_INDEXER_*`, `BROWSER_SESSIONS_*`, `EVALUATORS_*`, `PAYLOADS_*`, `TRACE_SUMMARY_*`, `NOTIFICATIONS_*`, `CLUSTERING_*`).
>     - Remove `worker_tracking` module and related health route; keep RabbitMQ-based `check_health`/`check_ready`.
>   - **Behavioral tweaks**:
>     - Standardized error handling with backoff; transient errors requeue (e.g., Quickwit ingest, clustering lock timeout).
>     - Span pipeline filters out non-recordable spans before ClickHouse insert; preserves notifications/evaluators/clustering flows.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=lmnr-ai%2Flmnr&utm_source=github&utm_medium=referral)<sup> for a0c2b601d23e209a85d1c96c98edaf549e8b8735. You can [customize](https://app.ellipsis.dev/lmnr-ai/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/app-server/src/main.rs
+++ b/app-server/src/main.rs
@@ -871,10 +871,12 @@ fn main() -> anyhow::Result<()> {
                     // Spawn notification workers
                     {
                         let db = db_for_consumer.clone();
+                        let client = reqwest::Client::new();
+
                         worker_pool_clone.spawn(
                             WorkerType::Notifications,
                             num_notification_workers as usize,
-                            move || NotificationHandler::new(db.clone(), reqwest::Client::new()),
+                            move || NotificationHandler::new(db.clone(), client.clone()),
                             QueueConfig {
                                 queue_name: NOTIFICATIONS_QUEUE,
                                 exchange_name: NOTIFICATIONS_EXCHANGE,
@@ -886,10 +888,11 @@ fn main() -> anyhow::Result<()> {
                     // Spawn clustering workers using new worker pool
                     {
                         let cache = cache_for_consumer.clone();
+                        let client = reqwest::Client::new();
                         worker_pool_clone.spawn(
                             WorkerType::Clustering,
                             num_clustering_workers as usize,
-                            move || ClusteringHandler::new(cache.clone(), reqwest::Client::new()),
+                            move || ClusteringHandler::new(cache.clone(), client.clone()),
                             QueueConfig {
                                 queue_name: CLUSTERING_QUEUE,
                                 exchange_name: CLUSTERING_EXCHANGE,
--- a/app-server/src/traces/consumer.rs
+++ b/app-server/src/traces/consumer.rs
@@ -266,15 +266,11 @@ async fn process_batch(
     }
 
     // Filter out spans that should not be recorded to clickhouse
-    let spans: Vec<Span> = spans
-        .into_iter()
-        .filter(|span| span.should_record_to_clickhouse())
-        .collect();
-
     let ch_spans: Vec<CHSpan> = spans
         .iter()
         .zip(span_usage_vec.iter())
         .zip(spans_ingested_bytes.iter())
+        .filter(|((span, _), _)| span.should_record_to_clickhouse())
         .map(|((span, span_usage), ingested_bytes)| {
             CHSpan::from_db_span(
                 &span,
@@ -297,6 +293,12 @@ async fn process_batch(
             e
         ));
     }
+
+    // Temporary solution to filter out spans before sending realtime span updates
+    let spans: Vec<Span> = spans
+        .into_iter()
+        .filter(|span| span.should_record_to_clickhouse())
+        .collect();
 
     // Send realtime span updates directly to SSE connections after successful ClickHouse writes
     send_span_updates(&spans, &pubsub).await;
--- a/app-server/src/traces/summary.rs
+++ b/app-server/src/traces/summary.rs
@@ -153,38 +153,35 @@ async fn process_trace_summary(
         )
         .await?;
 
-        if !channels.is_empty() {
-            // Push a notification for each configured channel
-            for channel in channels {
-                let payload = notifications::TraceAnalysisPayload {
-                    summary: response.summary.clone(),
-                    analysis: response.analysis.clone(),
-                    analysis_preview: response.analysis_preview.clone(),
-                    status: response.status.clone(),
-                    span_ids_map: response.span_ids_map.clone(),
-                    channel_id: channel.channel_id.clone(),
-                    integration_id: channel.integration_id,
-                };
+        // Push a notification for each configured channel
+        for channel in channels {
+            let payload = notifications::TraceAnalysisPayload {
+                summary: response.summary.clone(),
+                analysis: response.analysis.clone(),
+                analysis_preview: response.analysis_preview.clone(),
+                status: response.status.clone(),
+                span_ids_map: response.span_ids_map.clone(),
+                channel_id: channel.channel_id.clone(),
+                integration_id: channel.integration_id,
+            };
 
-                let notification_message = notifications::NotificationMessage {
-                    project_id: message.project_id,
-                    trace_id: message.trace_id,
-                    span_id: message.trigger_span_id,
-                    notification_type: NotificationType::Slack,
-                    event_name: event_name.to_string(),
-                    payload: serde_json::to_value(SlackMessagePayload::TraceAnalysis(payload))?,
-                };
+            let notification_message = notifications::NotificationMessage {
+                project_id: message.project_id,
+                trace_id: message.trace_id,
+                span_id: message.trigger_span_id,
+                notification_type: NotificationType::Slack,
+                event_name: event_name.to_string(),
+                payload: serde_json::to_value(SlackMessagePayload::TraceAnalysis(payload))?,
+            };
 
-                if let Err(e) =
-                    notifications::push_to_notification_queue(notification_message, queue.clone())
-                        .await
-                {
-                    log::error!(
-                        "Failed to push to notification queue for channel {}: {:?}",
-                        channel.channel_id,
-                        e
-                    );
-                }
+            if let Err(e) =
+                notifications::push_to_notification_queue(notification_message, queue.clone()).await
+            {
+                log::error!(
+                    "Failed to push to notification queue for channel {}: {:?}",
+                    channel.channel_id,
+                    e
+                );
             }
         }
     }
--- a/app-server/src/worker/mod.rs
+++ b/app-server/src/worker/mod.rs
@@ -201,8 +201,8 @@ impl<H: MessageHandler> QueueWorker<H> {
         let message = serde_json::from_slice::<H::Message>(data).map_err(|e| {
             log::error!(
                 "Queue message deserialization failed. Worker type: {:?}. Worker id: {}. Error: {:?}",
-                self.id,
                 self.worker_type,
+                self.id,
                 e
             );
             // Malformed message - reject without requeue (it won't deserialize on retry)


app-server/src/main.rs
```diff
@@ -732,156 +737,184 @@ fn main() -> anyhow::Result<()> {
...
                         App::new()
                             .wrap(NormalizePath::trim())
                             .app_data(web::Data::new(connection_for_health_clone.clone()))
-                            .app_data(web::Data::new(worker_tracker_clone.clone()))
-                            .app_data(web::Data::new(expected_counts.clone()))
+                            .app_data(web::Data::new(worker_pool_clone.clone()))
                             .app_data(web::Data::new(sse_connections.clone()))
                             .service(routes::probes::check_ready)
-                            .service(routes::probes::check_health_consumer)
+                            .service(routes::probes::check_health)
```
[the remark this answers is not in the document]
skull8888888: this is intended by design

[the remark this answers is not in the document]
dinmukhamedm: @skull8888888  this bit though
> The check_health_consumer function also becomes dead code.

Is this true? If so, let's remove the function

app-server/src/traces/clustering.rs
```diff
@@ -61,87 +58,36 @@ pub async fn push_to_clustering_queue(
...
+    fn on_error(&self, error: &anyhow::Error) -> crate::worker::ErrorAction {
+        let error_msg = error.to_string();
 
-        // Process the clustering message
-        if let Err(e) =
-            process_single_clustering(&client, cache.clone(), clustering_message, acker).await
-        {
-            log::error!("Failed to process clustering: {:?}", e);
+        // Requeue on lock timeout - another worker might get the lock next time
+        if error_msg.contains("Lock timeout") {
```
dinmukhamedm: the string comparison check by error message looks quite bad to me. Ideally we'd need some kind of an Error trait on the queue interface. And then the signature of `on_error` will change to accept a dynamic trait impl or a Box<dyn OurErrorTrait`

My idea, roughly:

```
pub trait QueueHandlerError {
    fn is_retriable(&self) {
        false
    }
}

struct ClusteringWorkerError {
    retriable: bool,
    // can also do:
    inner: anyhow::Error
}

impl QueueHandlerError for ClusteringWorkerError {
    fn is_retriable(&self) {
        self.retriable
    }
}

impl ClusteringWorkerError {
    fn from_anyhow(e: anyhow::Error, retriable: bool) -> Self {
        Self {
             inner: e
             retriable
        }
    }
}
```

dinmukhamedm: Another viable alternative is to use an enum, e.g. `enum QueueHandlerError` with the default variant being an anyhow error. With this we can benefit a lot from `thiserror` crate and its macros

app-server/src/traces/clustering.rs
```diff
@@ -153,12 +99,10 @@ async fn process_single_clustering(
...
         if start_time.elapsed() >= max_wait_duration {
             log::warn!(
-                "Timeout waiting for clustering lock for project_id={}, requeuing message",
+                "Timeout waiting for clustering lock for project_id={}, will retry",
                 message.project_id
             );
-            // Requeue the message to try again later
-            let _ = acker.reject(true).await; // true = requeue
-            return Ok(());
+            return Err(anyhow::anyhow!("Lock timeout")); // Will cause message to be requeued
```
dinmukhamedm: and here, we'd create the retriable variant of the error

app-server/src/traces/clustering.rs
```diff
@@ -177,17 +121,15 @@ async fn process_single_clustering(
...
-                if let Err(e) = acker.ack().await {
-                    log::error!("Failed to ack clustering message: {:?}", e);
-                }
-                return Ok(());
+                return Err(e.into());
             }
         }
     }
 
+    let client = reqwest::Client::new();
```
dinmukhamedm: +1 on bugbot's comment, should be injected from outside if possible

app-server/src/traces/clustering.rs
```diff
@@ -223,11 +165,9 @@ async fn process_single_clustering(
                 message.project_id,
                 e
             );
-            let _ = acker.ack().await;
+            Err(e)
```
dinmukhamedm: this used to `ack()`, now it does `reject.false()`. I am not exactly sure what the difference is, but let's double-check that this doesn't cause bad side effects

app-server/src/traces/consumer.rs
```diff
@@ -389,10 +309,6 @@ async fn process_batch(
         );
     }
 
-    let _ = acker.ack().await.map_err(|e| {
-        log::error!("Failed to ack MQ delivery (batch): {:?}", e);
-    });
-
     // Populate autocomplete cache
```
dinmukhamedm, line 396: there is a reason why the acknowledgment was placed here. The 100 lines worth of operations below (we used to agree, happy to re-discuss now) were non-critical to the span recording flow and would happen, so to speak, in the background.

To not change the current behaviour, we'll have to either `tokio::spawn` the rest of this function, or change the queue interface to have something like an `after_ack` callback.

Both options are not super nice in terms of code organization, so as I said, happy to re-discuss if that's needed at all, or we should only ack at the very end

app-server/src/traces/summary.rs
```diff
@@ -191,145 +113,91 @@ async fn process_trace_summary(
...
+    if let Some(event_name) = event_name {
+        // Get all channels configured for this event
+        let channels = db::slack_channel_to_events::get_channels_for_event(
+            &db.pool,
+            message.project_id,
+            event_name,
+        )
+        .await?;
+
+        if !channels.is_empty() {
```
dinmukhamedm: nit: now that we removed the debug log in the opposite case, this if check is not needed

app-server/src/worker/mod.rs
```diff
@@ -0,0 +1,274 @@
...
+}
+
+/// Queue worker that processes messages indefinitely
+pub struct QueueWorker<H: MessageHandler> {
+    id: Uuid,
+    worker_type: WorkerType,
+    handler: H,
+    queue: Arc<MessageQueue>,
+    config: QueueConfig,
+    state: Arc<RwLock<WorkerState>>,
```
dinmukhamedm: why is this needed? I don't think we ever read from that field

app-server/src/worker/mod.rs
```diff
@@ -0,0 +1,274 @@
...
+                    })
+            }
+        })
+        .await
+    }
+
+    async fn process_message(&self, data: &[u8]) -> Result<(), ErrorAction> {
+        let message = serde_json::from_slice::<H::Message>(data).map_err(|e| {
+            log::error!(
+                "Worker {} ({:?}) deserialization failed: {:?}",
```
dinmukhamedm: nit: I think this message is critical to track in logs. The filter regex to track it will be MUCH simpler, if we slightly reorder things here like:

```
Queue message deserialization failed. Worker type: {:?}. Worker id: {}. Error: {:?}
self.worker_type,
self.id,
e
```

app-server/src/worker/mod.rs
```diff
@@ -0,0 +1,274 @@
...
+    /// Spawn N workers of a type
+    pub fn spawn<H, F>(
+        &self,
+        worker_type: WorkerType,
+        count: usize,
+        handler_factory: F,
+        config: QueueConfig,
+    ) where
+        H: MessageHandler,
+        F: Fn() -> H + Send + Sync + 'static,
```
dinmukhamedm, line 246: interesting pattern. Looks a bit non-Rust, but I am fine. I guess the Rust way is to move that "factory" method onto the `MessageHandler` trait and implement on each handler instead of the `main.rs`, but I am not exactly sure what's the right way of configuring it from main then.

Not critical, but if you can quickly vibe-code around a more idiomatic way (possibly can avoid Send + Sync + 'static then), that will be great

app-server/src/routes/probes.rs
```diff
@@ -35,58 +34,3 @@ pub async fn check_ready(connection: web::Data<Option<Arc<Connection>>>) -> impl
 
     HttpResponse::Ok().body("OK")
 }
-
-#[get("/health")]
```
dinmukhamedm, line 39: Make sure to remove this from k8s config when releasing. Otherwise, this will respond with 404 and kube will keep restarting the consumer

dinmukhamedm: nvm, noticed that you didn't remove from main, just replaced with other route

app-server/src/traces/summary.rs
skull8888888: not sure what you mean here?

app-server/src/traces/clustering.rs
skull8888888: right now error rejects without requeueing

app-server/src/traces/clustering.rs
```diff
@@ -177,17 +116,13 @@ async fn process_single_clustering(
             }
             Err(e) => {
                 log::error!("Failed to acquire clustering lock: {:?}", e);
-                // Log and continue without clustering
-                if let Err(e) = acker.ack().await {
-                    log::error!("Failed to ack clustering message: {:?}", e);
-                }
-                return Ok(());
+                return Err(HandlerError::permanent(e));
```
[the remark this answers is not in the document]
dinmukhamedm: HandlerError::permanent results in `requeue: false`, so the DLX comment is irrelevant. The semantics does change, but that's somewhat intentional

app-server/src/traces/consumer.rs
dinmukhamedm: Discussed offline that this was not actually making anything quicker and that initially the code was written when our rabbitmq wasn't using the disk properly and had memory issues. Now the code and the infra is optimized better and this pattern is a bit less relevant

dinmukhamedm (approved):

PR #1126, opened 2026-01-06
Hybrid

Rainhunter13:
New backend only PR: https://github.com/lmnr-ai/lmnr/pull/1129

DONE:
- clickhouse spans/traces/tags writes to data plane
- clickhouse reads from data plane
- storage writes/reads to/from data plane

UI:
- [x] Disable "Save configuration" button unless keys are generated and url is verified
- [x] Confirmation window when changing the mode

ISSUES TO FIX:
- [ ]  ❗ Next.js is requesting clickhouse directly and all the routing on the app-server side is skipped (e.g. get span tags).   Options:
(A) Copy the routing and data plane calling logic to the Next.js. (-) duplicate logic, hard to sync changes 
(B) Make all clickhouse queries via app-server endpoint. (-) extra call to app-server (negligible?) ⭐ 
(C) Do routing only on Next, if Cloud then use clickhouse client, if Hybrid then call app-server (-) a bit messy, still extra app-server call for Hybrid mode
   *Actually, why some queries are done directly to Clickhouse and some via app-server? e.g. get spans vs get span tags
- [x] [FE] Invalidate cached config after keys generate?  UPD: no need, should invalidate only once the "mode" is switched
- [x] [FE] Db update check if no deployment exists.

TODO (important):
- [ ] [Terraform] Add S3 config to the env file
- [ ] [FE] Add "add-on" option to Pro Tier for hybrid mode
- [ ] [FE] Validate data plane url before making a test call
- [ ] S3 Cleanup

NEXT STEPS:
- [ ] Clickhouse writes for other types: evaluations, datasets, etc.
- [ ] [Data Plane] Implementation of Storage with GCS (Google Cloud Storage)

ASSUMPTIONS:
- Generating keys allowed only once (otherwise, need to invalidate cached config and cached auth token which is currently implemented using in-mem cache)

QUESTIONS:
- Storage: 
  - [ ] During S3 stream errors silently converted to empty bytes, shouldn't we return error instead? => Probably a bug
  - [ ] Why pushing to the queue is implemented at the Storage level? Shouldn't that behaviour be same no matter the storage implementation? => Probably was quickest implementation
  - [x] To enable reads from data plane S3, now all S3 reads are done in the app-server. Didn't do routing at the Next level for simplicity. Sounds ok? => Ok
  - [x] Is payloads api (for use outside of Next) is used anywhere? sdk? => BrowserUse use this api directly

Previous PR (same branch, just too much outdated comments):
- https://github.com/lmnr-ai/lmnr/pull/1073

### DIFF
--- a/frontend/components/workspace/deployment-settings/workspace-deployment.tsx
+++ b/frontend/components/workspace/deployment-settings/workspace-deployment.tsx
@@ -139,7 +139,7 @@ const WorkspaceDeployment = () => {
   if (isLoading) {
     return (
       <>
-        <SettingsSectionHeader title="Deployment" description="Choose how your application is deployed" />
+        <SettingsSectionHeader title="Deployment" description="Choose how your application is deployed." />
         <SettingsSection>
           <div className="flex flex-col gap-4">
             <Skeleton className="h-16 w-full" />
@@ -175,7 +175,14 @@ const WorkspaceDeployment = () => {
             <div className="flex items-start space-x-3 border rounded-lg p-4 cursor-pointer hover:bg-accent">
               <RadioGroupItem value={DeploymentType.CLOUD} id="cloud" />
               <Label htmlFor="cloud" className="flex-1 cursor-pointer">
-                <div className="font-medium">Cloud Deployment</div>
+                <div className="font-medium flex items-center gap-2">
+                  Cloud Deployment
+                  {data?.mode === DeploymentType.CLOUD && (
+                    <span className="text-xs bg-primary/10 text-primary px-2 py-0.5 rounded-full font-normal">
+                      Active
+                    </span>
+                  )}
+                </div>
                 <div className="text-sm text-muted-foreground">Fully managed hosting by Laminar.</div>
               </Label>
             </div>
@@ -183,9 +190,16 @@ const WorkspaceDeployment = () => {
             <div className="flex items-start space-x-3 border rounded-lg p-4 cursor-pointer hover:bg-accent">
               <RadioGroupItem value={DeploymentType.HYBRID} id="hybrid" />
               <Label htmlFor="hybrid" className="flex-1 cursor-pointer">
-                <div className="font-medium">Hybrid Deployment</div>
+                <div className="font-medium flex items-center gap-2">
+                  Hybrid Deployment
+                  {data?.mode === DeploymentType.HYBRID && (
+                    <span className="text-xs bg-primary/10 text-primary px-2 py-0.5 rounded-full font-normal">
+                      Active
+                    </span>
+                  )}
+                </div>
                 <div className="text-sm text-muted-foreground">
-                  Self-host your data while we manage the control plane.
+                  Self-host your data while we manage everything else.
                 </div>
               </Label>
             </div>
@@ -230,7 +244,6 @@ const WorkspaceDeployment = () => {
                     Warning: If your data plane is not configured properly, this may result in data loss or inability to
                     access your traces and logs.
                   </p>
-                  <p>Make sure you have properly set up and verified your data plane before proceeding.</p>
                 </>
               ) : (
                 <>


app-server/src/api/v1/evaluators.rs
```diff
@@ -56,16 +57,21 @@ pub enum CreateEvaluatorScoreRequest {
 #[post("/evaluators/score")]
 pub async fn create_evaluator_score(
     req: Json<CreateEvaluatorScoreRequest>,
-    db: Data<DB>,
+    db: web::Data<DB>,
```
dinmukhamedm, line 60: super nit: why? Data already imported. Same in lines 65-66

app-server/src/api/v1/payloads.rs
```diff
@@ -3,30 +3,26 @@ use std::sync::Arc;
...
 
-#[get("payloads/{payload_id}")]
-pub async fn get_payload(
-    path: web::Path<String>,
-    query: web::Query<PayloadQuery>,
-    storage: web::Data<Arc<Storage>>,
-    project_api_key: ProjectApiKey,
+/// Core logic for fetching a payload from storage.
+/// This is shared between the API key-authenticated endpoint and the project-scoped endpoint.
+pub async fn get_payload_response(
```
dinmukhamedm, line 20: nit, we usually move such functions into a separate module, e.g. `src/storage/payloads.rs` or `srs/payloads/mod.rs`, and import from there into both the API and frontend routes

app-server/src/ch/mod.rs
```diff
@@ -1,12 +1,64 @@
...
+        match self {
+            Table::Spans => "spans",
+            Table::Traces => "traces_replacing",
+            Table::Tags => "tags",
+        }
+    }
+}
+
+/// Trait for ClickHouse row types that can be inserted directly or via data plane.
+pub trait ClickhouseInsertable: RowOwned + RowWrite + Clone + Sized + Send + Sync {
```
dinmukhamedm, line 44: interesting, goes it not have to be `serde:Serialize`? Or is that guaranteed by `RowWrite`?

app-server/src/data_plane/auth.rs
```diff
@@ -0,0 +1,98 @@
...
+use super::crypto::decrypt_workspace_str;
+
+/// Token expiration time in seconds (15 minutes)
+const TOKEN_EXPIRATION_SECS: i64 = 900;
+
+/// Cache TTL - refresh token when 80% of lifetime has passed (12 minutes)
+const TOKEN_CACHE_TTL_SECS: u64 = 720;
+
+/// Cache of tokens per workspace_id
+static TOKEN_CACHE: OnceLock<Cache<Uuid, String>> = OnceLock::new();
```
dinmukhamedm, line 33: Why `moka::sync`? Can we use our `Cache` implementation? How does this work with the fact that app-server is on several machines?

Also, if after we discuss this and decide we still need a static OnceLock/LazyLock we need to triple check that it does what you want. From my quick experiments it wasn't as simple, and it seemed to recreate on every thread or something

app-server/src/data_plane/crypto.rs
```diff
@@ -0,0 +1,117 @@
...
+            "AEAD_SECRET_KEY must be 32 bytes (64 hex characters), got {} bytes",
+            key_bytes.len()
+        ));
+    }
+
+    xchacha20poly1305_ietf::Key::from_slice(&key_bytes)
+        .ok_or_else(|| anyhow!("Failed to create key from bytes"))
+}
+
+#[allow(dead_code)]
```
dinmukhamedm, line 24: what's the idea behind adding this function if it is unused? Also, should we possibly make this more generic and use the same interface in `src/notifications/slack.rs`? It's fine if not, and definitely out of scope of this PR, but if we do, it may make sense to make all interfaces in this file more general and place them in a separate module, e.g. `src/crypto`

app-server/src/data_plane/read.rs
```diff
@@ -0,0 +1,221 @@
...
+const DEFAULT_SQL_QUERY_MAX_RESULT_BYTES: &str = "536870912"; // 512MB
+
+#[derive(Serialize, Debug)]
+struct DataPlaneReadRequest {
+    query: String,
+    project_id: Uuid,
+    parameters: HashMap<String, Value>,
+}
+
+// TODO: move read() and read_from_clickhouse() to sql module, keep only data plane interactions here
```
dinmukhamedm, line 33: could you share more context on why you are keeping this TODO here? As in, is it deliberately out of scope of this PR, or have you just forgotten about this?

app-server/src/db/workspaces.rs
```diff
@@ -0,0 +1,51 @@
...
+    let result = sqlx::query_as::<_, WorkspaceDeployment>(
+        "
+        SELECT
+            projects.workspace_id,
+            COALESCE(workspace_deployments.mode, 'CLOUD') as mode,
+            COALESCE(workspace_deployments.public_key, '') as public_key,
+            COALESCE(workspace_deployments.private_key, '') as private_key,
+            COALESCE(workspace_deployments.private_key_nonce, '') as private_key_nonce,
+            COALESCE(workspace_deployments.data_plane_url, '') as data_plane_url,
+            COALESCE(workspace_deployments.data_plane_url_nonce, '') as data_plane_url_nonce
```
dinmukhamedm, line 39: why all this and not just typing them as `Option<String>` on the struct?

app-server/src/storage/data_plane.rs
```diff
@@ -0,0 +1,159 @@
...
+
+use crate::data_plane::{auth::generate_auth_token, crypto};
+use crate::db::workspaces::WorkspaceDeployment;
+
+/// Payload for storage upload requests to data plane.
+#[derive(Serialize)]
+struct StorageUploadPayload<'a> {
+    bucket: &'a str,
+    key: &'a str,
+    #[serde(with = "base64_serde")]
```
dinmukhamedm, line 20: is this the only reason why we are installing `serde_with` dependency? I think, this can be done with `serialize_with=base64_serde::serialize` and passing the serializer function directly. I might be slightly wrong with the details, but please have a look. We don't need a separate crate just for this

app-server/src/storage/data_plane.rs
```diff
@@ -0,0 +1,159 @@
...
+
+        let payload = StorageUploadPayload {
+            bucket,
+            key,
+            data: &data,
+        };
+
+        let response = self
+            .http_client
+            .post(format!("{}/api/v1/storage/upload", data_plane_url))
```
dinmukhamedm, line 88: super nit, but like 50 versions of Rust later LLMs still haven't figured out that you could do 
`format!("{data_plane_url}/api/v1/storage/upload")`, which IMO is much easier to read. Ditto below with bearer and in other methods. It's a personal preference though, so up to you

frontend/lib/crypto.ts
```diff
@@ -89,3 +89,67 @@ export async function encodeSlackToken(teamId: string, token: string): Promise<{
     throw new Error(`Failed to encode Slack token for team ${teamId}`);
   }
 }
+
+export async function encryptValue(
```
dinmukhamedm, line 93: same as my comment above on a similar function in app-server. Is the idea here that this is a more generic version of the *identical* function `encodeApiKey` and that we will gradually migrate to this, but it's out of scope of this PR? If so, let's add an inline TODO comment here and where `encodeApiKey` and `encodeSlackToken` are called.

PR #1132, opened 2026-01-11
insert items to labeling queue endpoint

Rainhunter13:
> [!NOTE]
> Introduces bulk ingestion for labeling queues and makes queue navigation deterministic across API, DB, and UI.
> 
> - **API/DB**: Adds `POST /v1/labeling_queues/{queue_id}/items` to insert multiple items; validates project ownership via `queue_exists`; batch inserts with `insert_labeling_queue_items`; wires route under `/v1/labeling_queues`.
> - **Queue navigation**: Updates move logic to order by `(created_at, id)` and requires `refId`; first-load uses epoch and empty UUID.
> - **UI/Flows**: `queue.tsx` displays source link when available or "Created manually"; navigation/actions send `refId`; `AddToLabelingQueuePopover`/`span-controls` pass optional `traceId`; makes push item metadata fields optional in schemas.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f7b8929297e81b1995e11bdc8b376fe2e87ea418. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

----

> [!IMPORTANT]
> Adds batch insertion endpoint for labeling queues, updates database logic, and modifies frontend for enhanced queue navigation and source context display.
> 
>   - **API/Server**:
>     - Adds `POST /v1/labeling_queues/{queue_id}/items` in `labeling_queues.rs` to insert multiple items into a labeling queue.
>     - Validates queue ownership and returns created `id` and `createdAt`.
>     - Registers route in `main.rs` under `/v1/labeling_queues` with ingestion auth.
>   - **Database**:
>     - New `labeling_queues.rs` module with `queue_exists` and `insert_labeling_queue_items()` for batch insertion using `UNNEST`.
>     - Updates `mod.rs` to include `labeling_queues` module.
>   - **Frontend**:
>     - Updates `queue.tsx` to use compound ordering `(createdAt, id)` for queue navigation.
>     - Modifies `move` API route and hotkeys to include `refId`.
>     - Displays source link or "Created manually" in `queue.tsx`.
>     - Updates `AddToLabelingQueuePopover` and `SpanControls` to pass `traceId`.
>     - Makes `metadata.source` and `id` optional in schemas in `index.ts`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=lmnr-ai%2Flmnr&utm_source=github&utm_medium=referral)<sup> for f7b8929297e81b1995e11bdc8b376fe2e87ea418. You can [customize](https://app.ellipsis.dev/lmnr-ai/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/app-server/src/db/labeling_queues.rs
+++ b/app-server/src/db/labeling_queues.rs
@@ -17,6 +17,7 @@ pub struct LabelingQueueItem {
 #[derive(Debug, Clone, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct NewLabelingQueueItem {
+    pub id: Uuid,
     pub metadata: serde_json::Value,
     pub payload: serde_json::Value,
 }
@@ -45,19 +46,26 @@ pub async fn insert_labeling_queue_items(
         return Ok(vec![]);
     }
 
-    let (metadatas, payloads): (Vec<serde_json::Value>, Vec<serde_json::Value>) = items
-        .into_iter()
-        .map(|item| (item.metadata, item.payload))
-        .unzip();
+    let len = items.len();
+    let mut ids = Vec::with_capacity(len);
+    let mut metadatas = Vec::with_capacity(len);
+    let mut payloads = Vec::with_capacity(len);
+
+    for item in items {
+        ids.push(item.id);
+        metadatas.push(item.metadata);
+        payloads.push(item.payload);
+    }
 
     let created_items = sqlx::query_as::<_, LabelingQueueItem>(
         r#"
-        INSERT INTO labeling_queue_items (queue_id, metadata, payload)
-        SELECT $1, metadata, payload FROM UNNEST($2::jsonb[], $3::jsonb[]) AS t(metadata, payload)
+        INSERT INTO labeling_queue_items (id, queue_id, metadata, payload)
+        SELECT id, $1, metadata, payload FROM UNNEST($2::uuid[], $3::jsonb[], $4::jsonb[]) AS t(id, metadata, payload)
         RETURNING id, created_at, queue_id, metadata, payload
         "#,
     )
     .bind(&queue_id)
+    .bind(&ids)
     .bind(&metadatas)
     .bind(&payloads)
     .fetch_all(pool)
--- a/app-server/src/db/labeling_queues.rs
+++ b/app-server/src/db/labeling_queues.rs
@@ -14,6 +14,13 @@ pub struct LabelingQueueItem {
     pub payload: serde_json::Value,
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct NewLabelingQueueItem {
+    pub metadata: serde_json::Value,
+    pub payload: serde_json::Value,
+}
+
 /// Check if a labeling queue exists and belongs to the given project.
 pub async fn queue_exists(pool: &PgPool, queue_id: Uuid, project_id: Uuid) -> Result<bool> {
     let exists = sqlx::query_scalar::<_, bool>(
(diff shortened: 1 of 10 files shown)


app-server/src/db/labeling_queues.rs
```diff
@@ -0,0 +1,65 @@
...
+        INSERT INTO labeling_queue_items (queue_id, metadata, payload)
+        SELECT * FROM UNNEST($1::uuid[], $2::jsonb[], $3::jsonb[])
+        RETURNING id, created_at, queue_id, metadata, payload
+        "#,
+    )
+    .bind(&queue_ids)
+    .bind(&metadatas)
+    .bind(&payloads)
+    .fetch_all(pool)
+    .await?;
```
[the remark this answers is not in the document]
Rainhunter13: Thinking this would not be needed at the client side and can add unnecessary latency, so for now would keep it unordered with a comment in docs. Especially given there are no other apis to use those ids.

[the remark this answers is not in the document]
dinmukhamedm: @Rainhunter13 if we use ordering `(created_at, id)` in the frontend, can't we assign uuidv7 here explicitly? I think currently postgres assigns an id using `gen_random_uuid()` since we don't supply anything. But we could also supply it, so we can just do `Uuid::now_v7()`. This way, I think the fetching ordering is guaranteed to be the same as what's been sent to us.

I know it's not exactly what bugbot is pointing out, but I think still worth looking at.

app-server/src/db/labeling_queues.rs
```diff
@@ -0,0 +1,65 @@
...
+        metadatas.push(metadata);
+        payloads.push(payload);
+    }
+
+    let created_items = sqlx::query_as::<_, LabelingQueueItem>(
+        r#"
+        INSERT INTO labeling_queue_items (queue_id, metadata, payload)
+        SELECT * FROM UNNEST($1::uuid[], $2::jsonb[], $3::jsonb[])
+        RETURNING id, created_at, queue_id, metadata, payload
+        "#,
```
dinmukhamedm: 1. No need to unnest and thus insert into vectors the queue_id. I don't remember the exact syntax, but roughly
```sql
INSERT INTO labeling_queue_items (queue_id, metadata, payload)
SELECT $1, metadata, payload FROM UNNEST($2::jsonb[], $3::jsonb[]) as any_tmp_table_name(metadata, payload)
RETURNING id, created_at, queue_id, metadata, payload
```

2. Not a fan of tuples, especially if items have the same type. If you can see yourself adding a comment `// (metadata, payload)`, then (in Rust) it means there is a concept to incorporate that. Let's move to structs. This will also help us, if we decide to add `id` like I discuss in my comment below.

3. If you can avoid `let mut` Vecs and pushing through for loops, do so. Even in Python, it's considered more elegant to do list comprehension. Probably an unzip would do. It might be a bit clunky for 3 vectors, but still much cleaner and perhaps more efficient than this.

frontend/components/queue/queue.tsx
```diff
@@ -189,7 +202,8 @@ function QueueInner() {
   );
 
   useEffect(() => {
-    move(new Date(0).toISOString(), "next", "first-load");
+    // Use epoch date and empty UUID to get the first item
+    move(new Date(0).toISOString(), "00000000-0000-0000-0000-000000000000", "next", "first-load");
```
dinmukhamedm, line 206: this looks hacky to me, but at least couldn't we make the first 2 parameters `string | null`? so that we don't have to craft a nil UUID. cc @olzhik11

dinmukhamedm (changes requested):

frontend/components/queue/queue.tsx
olzhik11: we can make them optional, this would require separate handling on action level, but i think its fine to leave it this way

dinmukhamedm (approved):

PR #1153, opened 2026-01-26
Feat/retroactive analysis test

Rainhunter13:
> [!NOTE]
> Introduces a new "Signals" trace-analysis pipeline with Gemini batch processing, replacing the prior semantic-events flow end-to-end.
> 
> - Adds Signals workers (submission, pending, waiting) with new RabbitMQ exchanges/queues and per-message TTL; maps `SIGNALS_*` to existing semantic queue names for compatibility
> - New ClickHouse models and ops: `signal_events`, `signal_runs`, `signal_run_messages`; enriches `traces` with `span_names` and in-memory filter evaluation for trigger matching
> - New DB modules: `signals`, `signal_triggers` (with parsed filters), `signal_jobs`; removes legacy semantic-event modules and event definition cache path
> - Adds `POST /api/v1/projects/{project_id}/signal-job` to submit jobs; integrates internal tracing spans
> - Integrates Gemini client and batch APIs; prompts/tools for identification; end-to-end run lifecycle (messages, tool calls, event creation, notifications, clustering)
> - MQ publish API gains optional TTL; all call sites updated; notifications payload simplified (no span_id)
> - Feature flags updated (`Signals`, `TraceAnalysis`); various wiring in main for workers, queues, and health
> - Misc: `serde_yaml` added; CH traces aggregation collects unique span names; Slack message links point to trace only
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e8b7f89ab6a4024b63c945a33d3dd721786f8e7f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


app-server/src/cache/keys.rs
```diff
@@ -6,8 +6,9 @@ pub const PROJECT_API_KEY_CACHE_KEY: &str = "project_api_key";
 pub const PROJECT_CACHE_KEY: &str = "project";
 pub const WORKSPACE_LIMITS_CACHE_KEY: &str = "workspace_limits";
 pub const PROJECT_EVALUATORS_BY_PATH_CACHE_KEY: &str = "project_evaluators_by_path";
-pub const SEMANTIC_EVENT_TRIGGER_SPANS_CACHE_KEY: &str = "semantic_event_trigger_spans";
-pub const PROJECT_EVENT_NAMES_CACHE_KEY: &str = "project_event_names";
+pub const SIGNAL_TRIGGERS_CACHE_KEY: &str = "signal_triggers";
+pub const SIGNAL_TRIGGER_LOCK_CACHE_KEY: &str = "signal_trigger_lock";
+pub const SIGNAL_TRIGGER_LOCK_TTL_SECONDS: u64 = 7200;
```
dinmukhamedm: nit: this shouldn't be here, must be in the file where it is used only

app-server/src/ch/signal_run_messages.rs
```diff
@@ -0,0 +1,108 @@
...
+
+    // Build comma-separated list of quoted UUIDs
+    let run_ids_str = run_ids
+        .iter()
+        .map(|id| format!("'{}'", id))
+        .collect::<Vec<_>>()
+        .join(", ");
+
+    let query = format!(
+        "DELETE FROM signal_run_messages WHERE project_id = ? AND run_id IN ({})",
```
dinmukhamedm, line 101: was this tested against our Prod? I think we don't have lightweight deletes enabled on our production CH instance.

Also, nit: I know &[Uuid] is safe and strongly typed, but crafting a query like this looks a little bit hacky. Is there no better binding option?

app-server/src/db/signal_triggers.rs
```diff
@@ -0,0 +1,81 @@
...
+                Ok(f) => f,
+                Err(e) => {
+                    log::warn!(
+                        "Failed to parse filters for signal '{}': {:?}",
+                        db_trigger.signal_name,
+                        e
+                    );
+                    return None;
+                }
+            };
```
dinmukhamedm, line 67: nit: not certain this filtering logic should be in the DB function, but I am open to discussing this. I think the idea for the `src/db` module is that it's just the interfaces to the DB

app-server/src/db/trace.rs
```diff
@@ -128,6 +224,13 @@ pub async fn upsert_trace_statistics_batch(
     let mut traces = Vec::new();
 
     for agg in aggregations {
+        let span_names_jsonb: Value = agg
+            .span_names
+            .iter()
+            .map(|name| (name.clone(), Value::Bool(true)))
+            .collect::<serde_json::Map<String, Value>>()
+            .into();
```
dinmukhamedm, line 232: nit, let's add an inline comment explaining the choice of a map, not Array here. Something like

```sql
`SELECT '["a", "b"]'::jsonb || '["b", "c"]'::jsonb` returns `["a", "b", "b", "c"]`
but correctly merges objects: `SELECT '{"a": true, "b": true}'::jsonb || '{"b": true, "c": true}'::jsonb` returns `{"a": true, "b": true, "c": true}`
```

app-server/src/features/mod.rs
```diff
@@ -49,10 +50,11 @@ pub fn is_feature_enabled(feature: Feature) -> bool {
             env::var("CLUSTERING_SERVICE_URL").is_ok()
                 && env::var("CLUSTERING_SERVICE_SECRET_KEY").is_ok()
         }
-        Feature::SemanticEvents => {
+        Feature::Signals => {
             env::var("SEMANTIC_EVENT_SERVICE_URL").is_ok()
                 && env::var("SEMANTIC_EVENT_SERVICE_SECRET_KEY").is_ok()
         }
+        Feature::TraceAnalysis => env::var("GOOGLE_GENERATIVE_AI_API_KEY").is_ok(),
```
dinmukhamedm, line 57: shouldn't we have a separate key for this feature?

app-server/src/signals/queue.rs
```diff
@@ -0,0 +1,160 @@
...
+            log::warn!(
+                "[SIGNAL JOB] MQ payload limit exceeded. Project ID: [{}], Job ID: [{}], payload size: [{}]. Batch size: [{}]",
+                batch_message.project_id,
+                batch_message.job_id,
+                serialized.len(),
+                runs_batch.len()
+            );
+            // Skip publishing this batch
+            continue;
+        }
```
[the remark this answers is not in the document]
olzhik11: Batch size is configured on our side, and given our message structure, it can't it limit.

app-server/src/signals/gemini/client.rs
```diff
@@ -0,0 +1,98 @@
...
+pub type GeminiResult<T> = Result<T, GeminiError>;
+
+impl GeminiClient {
+    pub fn new() -> GeminiResult<Self> {
+        let api_key = env::var("GOOGLE_GENERATIVE_AI_API_KEY").map_err(|_| {
+            GeminiError::config("GOOGLE_GENERATIVE_AI_API_KEY environment variable not set")
+        })?;
+
+        let api_base_url = env::var("GEMINI_API_BASE_URL")
+            .unwrap_or_else(|_| "https://generativelanguage.googleapis.com/v1beta".to_string());
```
dinmukhamedm: nit: tab and claude will probably never learn this, but:
```rust
unwrap_or("https://generativelanguage.googleapis.com/v1beta".to_string());
```

app-server/src/signals/gemini/utils.rs
```diff
@@ -0,0 +1,94 @@
...
+        .map(|c| serde_json::to_string(&c.content).unwrap_or_default())
+        .unwrap_or_default();
+
+    let function_call =
+        candidate.and_then(|c| c.content.parts.iter().find_map(|p| p.function_call.clone()));
+
+    let text = candidate.and_then(|c| c.content.parts.iter().find_map(|p| p.text.clone()));
+
+    // Include thoughts tokens in output tokens.
+    // Divide by 2 with ceiling to account for discounted batching price.
```
dinmukhamedm, line 50: ❤️

app-server/src/signals/pendings_consumer.rs
```diff
@@ -0,0 +1,771 @@
...
+/// Handle succeeded batch
+async fn process_succeeded_batch(
+    message: &SignalJobPendingBatchMessage,
+    response: Option<GenerateContentBatchOutput>,
+    db: Arc<DB>,
+    queue: Arc<MessageQueue>,
+    clickhouse: clickhouse::Client,
+    config: Arc<SignalWorkerConfig>,
+) -> Result<(), HandlerError> {
+    let response = response.ok_or_else(|| {
```
dinmukhamedm: ditto.

```rust
    response.ok_or(
        HandlerError::permanent(anyhow::anyhow!(
            "Batch succeeded but response is missing for batch_id: {}",
            message.batch_id
        ))
    )?
```

app-server/src/signals/utils.rs
```diff
@@ -0,0 +1,214 @@
...
+    span_type: SpanType,
+    start_time: chrono::DateTime<Utc>,
+    input: Option<Value>,
+    output: Option<Value>,
+    input_tokens: Option<i32>,
+    output_tokens: Option<i32>,
+    model: Option<String>,
+    provider: Option<String>,
+    queue: Arc<MessageQueue>,
+    internal_project_id: Option<Uuid>,
```
dinmukhamedm, line 112: nit/lint, and possibly out of scope, because a relatively large change. Calls to this function are practically unreadable, especially without inlay hints. The idiomatic rust way to do so is to create a struct that you can create outside with all the named fields, and pass it to this function, similar to typescript object args.

app-server/src/signals/tools.rs
```diff
@@ -0,0 +1,153 @@
...
+                        "items": {"type": "integer"},
+                        "description": "List of span IDs (sequential integers starting from 1) to fetch full information for."
+                    }
+                },
+                "required": ["span_ids"]
+            }),
+        },
+        FunctionDeclaration {
+            name: "submit_identification".to_string(),
+            description: "Submits the final identification result. Call this when you have determined whether the semantic event can be identified in the trace and have extracted the relevant data (if identified=true) or determined it cannot be found (if identified=false).".to_string(),
```
dinmukhamedm, line 58: nit: I'd rather have these func descriptions in a separate file, together with prompts, but probs out of scope of this PR too

app-server/src/signals/pendings_consumer.rs
```diff
@@ -0,0 +1,771 @@
...
+    delete_signal_run_messages(clickhouse, message.project_id, &finished_run_ids).await?;
+
+    // Update job stats
+    update_signal_job_stats(
+        &db.pool,
+        message.job_id,
+        succeeded_runs.len() as i32,
+        failed_runs.len() as i32,
+    )
+    .await?;
```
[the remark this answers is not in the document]
Rainhunter13: but that's why we are calling that action last, if something failed before this action should not happen twice

PR #1189, opened 2026-02-05
fix: realtime traces constantly scrolling to selected span

kolbeyang:
This was caused by a useEffect scrolling to the selected span for condensed timeline span click feature. 

The fix was to trigger this behavior only on selectedSpanIndex change instead of listening to both selected span and all spans. 

Listening to all spans caused scroll to behavior anytime a new span arrived.

I split the `scrollTo` `useEffect` into a `useMemo` and a `useEffect`. The array of all spans dependency was moved from the `useEffect` to the `useMemo` to avoid retriggering the `useEffect` whenever a new span arrives.

The same fix was applied four times on (Tree, Reader) x (Rollout, Trace view)

---

> [!NOTE]
> **Low Risk**
> Small, localized UI behavior change affecting only auto-scroll behavior in four virtualized components; main risk is missing an intended scroll in edge cases when selection or span ordering changes.
> 
> **Overview**
> Stops the trace and rollout session *List* and *Tree* views from constantly jumping back to the selected span during realtime updates.
> 
> This refactors the scroll-to-selected logic to compute `selectedSpanIndex` in a `useMemo` and run the `scrollToIndex` effect only when that index changes (and is valid), rather than re-triggering on incoming span array changes; null/undefined handling was tightened via `isNil`/`isNull`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 172170ece348944e993cb252de14608549a1007b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/frontend/components/rollout-sessions/rollout-session-view/list/index.tsx
+++ b/frontend/components/rollout-sessions/rollout-session-view/list/index.tsx
@@ -59,7 +59,7 @@ const List = ({ traceId, onSpanSelect }: ListProps) => {
       });
       return () => cancelAnimationFrame(rafId);
     }
-  }, [selectedSpanIndex, virtualizer]);
+  }, [selectedSpanIndex, virtualizer, isSpansLoading]);
 
   const items = virtualizer?.getVirtualItems() || [];
 
--- a/frontend/components/rollout-sessions/rollout-session-view/list/index.tsx
+++ b/frontend/components/rollout-sessions/rollout-session-view/list/index.tsx
@@ -1,5 +1,5 @@
 import { useVirtualizer } from "@tanstack/react-virtual";
-import { compact, isEmpty, times } from "lodash";
+import { compact, isEmpty, isNil, isNull, times } from "lodash";
 import { useParams } from "next/navigation";
 import React, { useCallback, useEffect, useMemo, useRef, useState } from "react";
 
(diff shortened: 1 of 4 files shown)


frontend/components/rollout-sessions/rollout-session-view/list/index.tsx
```diff
@@ -44,24 +44,22 @@ const List = ({ traceId, onSpanSelect }: ListProps) => {
...
-        virtualizer.scrollToIndex(selectedIndex, {
-          align: "start",
-        });
-      }
-    };
+  const selectedSpanIndex = useMemo(() => {
+    if (isNil(selectedSpan)) return null;
+    const selectedIndex = listSpans.findIndex((span) => span.spanId === selectedSpan.spanId);
+    return selectedIndex;
+  }, [selectedSpan?.spanId, listSpans, isSpansLoading]);
```
dinmukhamedm: super nit (I don't really understand react, so might be completely wrong): why do all `useMemo`s here depend on `isSpansLoading` even this one that doesn't use it?

dinmukhamedm (dismissed):

frontend/components/rollout-sessions/rollout-session-view/list/index.tsx
kolbeyang: You're right this is unneeded

dinmukhamedm (approved):

PR #1212, opened 2026-02-11
fix data count

skull8888888:
> [!NOTE]
> **Medium Risk**
> Changes affect how ingested byte counts are computed and stored, which can impact billing/usage limits and analytics if `estimate_size_bytes` is missed or computed at the wrong lifecycle point.
> 
> **Overview**
> Spans now persist a `size_bytes` field (`db::spans::Span`) and `estimate_size_bytes` was changed from a pure getter to a mutating method that populates this field immediately after span creation/deserialization.
> 
> ClickHouse serialization (`CHSpan`) and workspace usage-limit byte accounting in the spans consumer were updated to use `span.size_bytes` (computed once up front) rather than recomputing size later, and all span construction call sites/tests initialize `size_bytes` to `0` for backwards compatibility.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6bb6785e4888c846e38f5cd8704a5e742a5f105b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/app-server/src/signals/utils.rs
+++ b/app-server/src/signals/utils.rs
@@ -216,6 +216,7 @@ pub async fn emit_internal_span(queue: Arc<MessageQueue>, span: InternalSpan) ->
         tags: None,
         input_url: None,
         output_url: None,
+        size_bytes: 0,
     };
 
     if let Some(error) = span.error {
--- a/app-server/src/traces/spans.rs
+++ b/app-server/src/traces/spans.rs
@@ -898,9 +898,8 @@ impl Span {
         Ok(())
     }
 
-    pub fn estimate_size_bytes(&self) -> usize {
-        // events size is estimated separately, so ignored here
-
+    /// This function MUST to be called right after we deserialize or create a span object.
+    pub fn estimate_size_bytes(&mut self) {
         // 16 bytes for span_id,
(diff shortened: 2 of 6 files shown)


app-server/src/traces/spans.rs
```diff
@@ -898,17 +898,17 @@ impl Span {
         Ok(())
     }
 
-    pub fn estimate_size_bytes(&self) -> usize {
-        // events size is estimated separately, so ignored here
-
+    /// This function MUST to be called right after we deserialize or create a span object.
+    pub fn estimate_size_bytes(&mut self) {
```
Rainhunter13, line 902: may be call it `set_size_bytes` to somehow indicate we are mutating the object?

app-server/src/signals/utils.rs
```diff
@@ -216,6 +216,7 @@ pub async fn emit_internal_span(queue: Arc<MessageQueue>, span: InternalSpan) ->
         tags: None,
         input_url: None,
         output_url: None,
+        size_bytes: 0,
```
Rainhunter13, line 219: shall we call the `.estimate_size_bytes()` function somewhere there?

PR #1248, opened 2026-02-19
feat: add signal limits, small refactor

olzhik11:
> [!NOTE]
> **Medium Risk**
> Adds gating logic to a write-path endpoint and depends on cached/ClickHouse usage calculations; mistakes could incorrectly block or allow signal jobs on the free tier.
> 
> **Overview**
> Signal job creation now enforces **free-tier monthly signal-runs limits**: the `POST /api/projects/:projectId/signals/:id/jobs` route computes an expected run count (passed from the client) and calls a new `checkSignalRunsLimit` guard before queueing the job.
> 
> Adds `frontend/lib/actions/usage/limits.ts`, which looks up a project’s workspace tier/reset time (via cache/DB), reads usage (cache or ClickHouse since last reset), and throws a user-facing error when the requested runs would exceed the limit. Client-side job creation now includes `tracesCount` in the request and surfaces server error messages, and workspace usage/stats imports are refactored to `@/lib/actions/usage/*` paths.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2b34c6a9576f6db27df2e60dbe4c920f529575f0. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/frontend/lib/actions/usage/limits.ts
+++ b/frontend/lib/actions/usage/limits.ts
@@ -2,15 +2,46 @@ import { addMonths } from "date-fns";
 import { eq } from "drizzle-orm";
 
 import { completeMonthsElapsed } from "@/lib/actions/workspaces/utils";
-import { cache, WORKSPACE_SIGNAL_RUNS_USAGE_CACHE_KEY } from "@/lib/cache";
+import { cache, PROJECT_CACHE_KEY, WORKSPACE_SIGNAL_RUNS_USAGE_CACHE_KEY } from "@/lib/cache";
 import { clickhouseClient } from "@/lib/clickhouse/client";
 import { db } from "@/lib/db/drizzle";
 import { projects, subscriptionTiers, workspaces } from "@/lib/db/migrations/schema";
 import { Feature, isFeatureEnabled } from "@/lib/features/features";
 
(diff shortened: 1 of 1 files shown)


frontend/lib/actions/usage/limits.ts
```diff
@@ -0,0 +1,85 @@
...
+    .select({
+      workspaceId: workspaces.id,
+      signalRunsLimit: subscriptionTiers.signalRuns,
+      resetTime: workspaces.resetTime,
+    })
+    .from(projects)
+    .innerJoin(workspaces, eq(projects.workspaceId, workspaces.id))
+    .innerJoin(subscriptionTiers, eq(workspaces.tierId, subscriptionTiers.id))
+    .where(eq(projects.id, projectId))
+    .limit(1);
```
dinmukhamedm, line 58: There's a function in Rust `crate::db::projects::get_project_and_workspace_billing_info` that does exactly this (and is cached one layer above in `crate::utils::limits::get_workspace_info_for_project_id`). I don't think we have implemented it in Next, but maybe we should?

PR #1462, opened 2026-03-22
fetch session images on request only

Rainhunter13:
> [!NOTE]
> **Low Risk**
> Low risk UI/performance change that defers image fetching until the Images tab is opened; main risk is missing images if any flows relied on eager loading or on `output` being queried for images.
> 
> **Overview**
> **Span images are now fetched on-demand.** Both session player variants track whether the Images tab has ever been opened and pass `enabled` to `SpanImagesVideoPlayer` so SWR only requests/preloads images after the tab is selected (or immediately when there is no browser session).
> 
> **Backend query is slimmed down.** `getSpanImages` no longer selects/parses `output`, reducing the data pulled from ClickHouse for image extraction.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit cb7d8f95a0e94de0f2531db86068e5d0eb4d47d2. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/frontend/components/shared/traces/session-player.tsx
+++ b/frontend/components/shared/traces/session-player.tsx
@@ -49,6 +49,7 @@ const SessionPlayer = ({ hasBrowserSession, traceId, llmSpanIds = [], onClose }:
   const [currentUrl, setCurrentUrl] = useState("");
   const [urlChanges, setUrlChanges] = useState<UrlChange[]>([]);
   const [activeTab, setActiveTab] = useState(hasBrowserSession ? "browser-session" : "images");
+  const [imagesTabOpened, setImagesTabOpened] = useState(!hasBrowserSession);
   const [dimensions, setDimensions] = useState({ width: 800, height: 600 });
   const [speed, setSpeed] = useLocalStorage("session-player-speed", 1);
 
@@ -256,7 +257,10 @@ const SessionPlayer = ({ hasBrowserSession, traceId, llmSpanIds = [], onClose }:
         )}
 
         <button
-          onClick={() => setActiveTab("images")}
+          onClick={() => {
+            setActiveTab("images");
+            setImagesTabOpened(true);
+          }}
           className={`mx-2 inline-flex items-center justify-center whitespace-nowrap border-b-2 py-1.5 text-sm transition-all gap-2 font-medium ${
             activeTab === "images"
               ? "border-secondary-foreground text-foreground"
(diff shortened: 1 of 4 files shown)


frontend/components/traces/session-player.tsx
```diff
@@ -54,6 +54,7 @@ const SessionPlayer = ({ hasBrowserSession, traceId, llmSpanIds = [], onClose }:
   const [currentUrl, setCurrentUrl] = useState("");
   const [urlChanges, setUrlChanges] = useState<UrlChange[]>([]);
   const [activeTab, setActiveTab] = useState(hasBrowserSession ? "browser-session" : "images");
+  const [imagesTabOpened, setImagesTabOpened] = useState(!hasBrowserSession);
```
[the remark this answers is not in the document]
Rainhunter13: can't happen in practice

frontend/components/shared/traces/session-player.tsx
```diff
@@ -353,7 +357,7 @@ const SessionPlayer = ({ hasBrowserSession, traceId, llmSpanIds = [], onClose }:
         </div>
 
         <div className={`h-full ${activeTab === "images" ? "block" : "hidden"}`}>
-          <SpanImagesVideoPlayer traceId={traceId} spanIds={llmSpanIds} isShared />
+          <SpanImagesVideoPlayer traceId={traceId} spanIds={llmSpanIds} isShared enabled={imagesTabOpened} />
```
olzhik11, line 360: `enabled={activeTab === "images"}` ? we might not need additional state if i got it right

olzhik11: ~but even better i would just render it conditionally without prop:~
```
{activeTab === 'images' && <SpanImagesVideoPlayer ....
```

Rainhunter13: If you use enabled={activeTab === "images"} to avoid the extra state, then every time the user switches away from Images and back, SWR's data goes undefined → cached value, which triggers the images useMemo → triggers the useEffect → setSessionTime(0). That's not a genuine data change (no new spans arrived), it's just the same cached data being restored — but it still resets the browser session player.
The imagesTabOpened latch is already minimal — one boolean. It keeps the SWR key stable after the first open so the data never cycles through undefined, avoiding the spurious reset while still allowing real realtime updates to flow through 🤔

Issue #838, opened 2025-09-08
How lmnr caculate the token cost?


yunnysunny:
I use mastra to report my trace to lmnr. And I found all traces cost was zero.

<img width="987" height="744" alt="Image" src="https://github.com/user-attachments/assets/3e48f85e-c207-486d-93c5-6de2a44beed8" />

I think it maybe mastra's fault. But I'm not sure. In the above picture , the `Span Input` and `Span Output` is null, and the `Events` is `[]`, the Attributes is :
```yaml
componentName: My Agent
aisdk.model.id: glm-4.5-flash
lmnr.span.path:
  - mastra.stream
operation.name: mastra.stream
lmnr.span.ids_path: []
mastra.operationId: mastra.stream
aisdk.model.provider: openai.chat
stream.response.text: 你好！很高兴见到你。我是Claude，一个AI助手，可以帮助你回答问题、提供信息或进行对话。有什么我可以帮你的吗？
stream.prompt.messages: '[{"role":"user","content":[{"type":"text","text":"你好"}]}]'
stream.usage.inputTokens: 14
stream.usage.totalTokens: 101
stream.response.toolCalls: "[]"
stream.usage.outputTokens: 87
stream.settings.temperature: 0
stream.response.finishReason: stop
stream.usage.cachedInputTokens: 12
stream.response.providerMetadata: '{"openai":{}}'
```

yunnysunny:
I use mastra itselt to report traces, I can not change the way of reporting in mastra. 
The mastra.stream.aisdk.doStream span attributes:
```yaml
componentName: My Agent
aisdk.model.id: glm-4.5-flash
lmnr.span.path:
  - mastra.stream.aisdk.doStream
operation.name: mastra.stream.aisdk.doStream
lmnr.span.ids_path: []
mastra.operationId: mastra.stream.aisdk.doStream
aisdk.model.provider: openai.chat
stream.prompt.toolChoice: auto
stream.response.timestamp: 2025-09-08T12:56:44.566Z
stream.settings.temperature: 0
stream.response.msToFirstChunk: 1
```
The package versions:

- @ai-sdk/openai: 2.0.25
- @mastra/core: 0.16.0

Since mastra use streamVNext internal when I call generateVNext, it's same issue I encountered.

<img width="799" height="790" alt="Image" src="https://github.com/user-attachments/assets/be1033c8-84ce-49da-b011-e90efa8d09af" />

[...]

PR #1548, opened 2026-03-31
feat: add support for vercel/mastra sdks

PranshuSrivastava:
Fixes: #838 
Add support for Vercel/ Mastra SDKs via normalisation of raw attributes.

---

> [!NOTE]
> **Medium Risk**
> Touches LLM span attribute normalization and span-type detection, which can change how usage/cost and inputs/outputs are extracted for incoming traces. Risk is moderate because it’s additive and guarded (no overwrites), but it affects core telemetry parsing paths.
> 
> **Overview**
> Adds support for newer Vercel AI SDK/Mastra telemetry by **normalizing `aisdk.*` and operation-prefixed keys** (e.g. `stream.*`, `generateText.*`) into the existing `gen_ai.*`/`ai.*` attributes during `Span::parse_and_enrich_attributes`, including model/provider, token usage, and prompt/response fields.
> 
> Updates LLM span classification to recognize `aisdk.*`, filters out operation-prefixed prompt attributes from persisted raw attributes, and adds warning logs for inconsistent token breakdowns (cache tokens > total) and for spans where token usage exists but pricing/model info is missing. Includes a comprehensive new test suite covering normalization behavior and non-overwrite guarantees.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b3b85baa87bca4e5a2375b62b2308dc7e24687d2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/app-server/src/traces/provider/mod.rs
+++ b/app-server/src/traces/provider/mod.rs
@@ -29,5 +29,10 @@ fn is_ai_sdk_llm_span(span: &Span) -> bool {
                 .attributes
                 .raw_attributes
                 .get("ai.model.provider")
+                .is_some()
+            || span
+                .attributes
+                .raw_attributes
+                .get("aisdk.model.provider")
                 .is_some())
 }
--- a/app-server/src/traces/span_attributes.rs
+++ b/app-server/src/traces/span_attributes.rs
@@ -49,3 +49,7 @@ pub const OPENAI_RESPONSE_SERVICE_TIER: &str = "openai.response.service_tier";
 pub const OPENAI_REQUEST_SERVICE_TIER: &str = "openai.request.service_tier";
 pub const ANTHROPIC_RESPONSE_SERVICE_TIER: &str = "anthropic.response.service_tier";
 pub const ANTHROPIC_REQUEST_SERVICE_TIER: &str = "anthropic.request.service_tier";
+
+// Newer Vercel AI SDK / Mastra attributes
+pub const AISDK_MODEL_ID: &str = "aisdk.model.id";
+pub const AISDK_MODEL_PROVIDER: &str = "aisdk.model.provider";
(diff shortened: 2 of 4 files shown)


JiwaniZakir (commented):
The `"stream"` entry in `AISDK_OPERATION_PREFIXES` (in `spans.rs`) looks suspect — `streamText` and `streamObject` are already explicit entries, and there's no standalone `"stream"` operation in the Vercel AI SDK. Since `detect_aisdk_operation_prefix` iterates in order, a span with `"streamText.usage.inputTokens"` won't be affected (the key lookup is exact), but the bare `"stream"` prefix could match unintended spans in other instrumentation libraries that happen to emit attributes like `"stream.usage.inputTokens"`.

In `normalize_aisdk_attributes`, the provider normalization splits on `'.'` and lowercases — e.g., `"openai.chat"` → `"openai"` — but `.trim()` on the result of `.split('.').next()` is a no-op because split results from a non-whitespace delimiter won't have leading/trailing spaces. It's harmless but slightly misleading.

The new warning in `input_tokens()` guards with `total_input_tokens > 0`, which silently suppresses the warning when `total_input_tokens == 0` but `cache_total > 0` — that's also a semantically invalid state worth surfacing. Consider logging a warning whenever `cache_total > total_input_tokens` unconditionally (or separately handle the zero case).

Finally, `normalize_aisdk_attributes` is `pub` but the diff is truncated — it would be worth confirming this is called before `span_type()` resolution (line ~416), since the type inference in that block relies on `GEN_AI_SYSTEM` being present, and the normalization needs to run first to populate it from `aisdk.model.provider`.

dinmukhamedm:
@PranshuSrivastava thank you! Continued in #1595

PR #1554, opened 2026-03-31
add an info comment about enabling events manually

dinmukhamedm:
> [!NOTE]
> **Low Risk**
> Low risk: only adds a code comment and removes dead commented-out logging in the webhook default case, with no runtime behavior change.
> 
> **Overview**
> Adds an inline note in `frontend/app/api/webhook/stripe/route.ts` reminding developers that when new Stripe webhook `event.type`s are handled, they must also be enabled manually via Stripe Workbench.
> 
> Also removes the previously commented-out “Unhandled event type” log from the `default` branch, keeping behavior unchanged (still no-op on unknown events).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit fce3613b1d5576b7fb8e2677d8cd24e04c216a15. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/frontend/app/api/webhook/stripe/route.ts
+++ b/frontend/app/api/webhook/stripe/route.ts
@@ -180,9 +180,8 @@ export async function POST(req: NextRequest): Promise<Response> {
     case "customer.subscription.updated":
       await handleSubscriptionChange(event);
       break;
+    // NOTE: if adding new events here, don't forget to enable them via Stripe Workbench
     default:
-      // Unexpected event type
-      // console.log(`Stripe Webhook. Unhandled event type ${event.type}.`);
       break;
   }
   return new Response("Webhook received.", { status: 200 });


JiwaniZakir (commented):
The new comment in `route.ts` is a useful reminder, but placing it before the `default` case is a bit misleading — it reads as if the `default` branch itself requires Stripe Workbench action, when the instruction really applies to the `case` handlers above. Moving it to the top of the `switch` block or just above the first `case` would better communicate that the note applies to the entire event-handling pattern.

Removing the commented-out `console.log` is a clean-up, but the deleted `// Unexpected event type` comment was doing some useful documentation work — without it (or any logging), silently swallowing unhandled event types makes it harder to notice when Stripe sends something unexpected or when a misconfigured webhook delivers events that never get handled. Even a no-op `default` could benefit from a brief note explaining the intentional silence (e.g., `// All other event types are intentionally ignored`).

It's also worth considering whether the Stripe Workbench note should link to or reference the specific dashboard section or configuration file, since "Stripe Workbench" is a broad term and a future developer may not know exactly where to look.

olzhik11 (approved):

PR #1579, opened 2026-04-03
fix: make email whitelist comparison case-insensitive

xingyaoww:
GitHub OAuth may return emails with different casing than what is configured in allowed-emails.json (e.g. 'User@gmail.com' vs 'user@gmail.com'). The previous list.includes() check is case-sensitive, causing legitimate users to be denied access.

Use case-insensitive comparison instead.

---

> [!NOTE]
> **Low Risk**
> Low risk, small change limited to the GitHub sign-in whitelist comparison logic. Main risk is unintentionally allowing emails that differ only by case from the configured list (typically desired).
> 
> **Overview**
> Updates the NextAuth `signIn` callback to perform a **case-insensitive** comparison when checking GitHub-authenticated users against `allowed-emails.json`, replacing `list.includes(user.email)` with a lowercased `some()` match.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 24b62265fa103a1ff4218d37c76a6311a07d09fe. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


dinmukhamedm:
NextAuth to BetterAuth migration removed this functionality completely. The recommended safe way to solve this when self-hosting is to use an external auth provider like keycloak or okta

PR #1590, opened 2026-04-06
fix: defer report scheduler checkpoint until enqueue succeeds

### LINKED ISSUE #1586: Reports scheduler advances checkpoint before enqueue completes, causing missed triggers on failure
I think there’s a subtle but serious issue in the reports scheduler around how the checkpoint (`REPORT_SCHEDULER_LAST_CHECK_CACHE_KEY`) is handled.

Right now in `check_and_enqueue`, the scheduler updates the checkpoint to `now` *before* it actually enqueues any report trigger messages. This effectively marks the entire time window as “processed” even though the enqueue step hasn’t completed yet.

That becomes a problem in failure scenarios:

* If the process crashes (or restarts) after the checkpoint is written but before the enqueue loop finishes, the next run will start from the new checkpoint and skip all the pending hour buckets in that window.
* If the message queue is unavailable (e.g. RabbitMQ down), `push_to_reports_queue` can fail for all reports, but the function still returns `Ok(())` and the checkpoint has already moved forward — so those triggers are never retried.
* Errors from enqueue are only logged, and the result of `cache.insert(...)` is ignored (`let _ =`), so there’s no strong signal that anything went wrong.

In practice this can lead to silently missing scheduled reports (daily/weekly), which is pretty hard to detect unless someone notices the absence of emails.

**Repro idea:**

* Let the scheduler run with at least one report due
* Kill the process right after the checkpoint is written (before enqueue completes), or simulate MQ failure
* Restart and observe that the missed time buckets are not retried

**Root cause (as I understand it):**
The checkpoint is being treated as “we attempted this window” instead of “we successfully handed off the work”.

**Possible fix directions:**

* Move the checkpoint update to *after* successful enqueue
* Or advance it incrementally (per hour bucket / per successful enqueue)
* Avoid ignoring errors from `cache.insert`
* Optionally add retry/backoff for queue failures

Happy to take a shot at a fix if this direction makes sense 👍

ixchio:
## Summary

Fixes #1586 — the reports scheduler was advancing its checkpoint **before** the enqueue loop completed, causing silently missed report triggers on process crash or MQ failure.

## Problem

In `check_and_enqueue`, the `REPORT_SCHEDULER_LAST_CHECK_CACHE_KEY` was updated to `now` at the top of the function (line 99), before any messages were actually published to the queue. This created a race condition:

1. **Process crash mid-enqueue**: If the process crashed after the checkpoint write but before all reports were enqueued, the next run would start from the already-advanced checkpoint — permanently skipping the pending hour buckets.
2. **MQ unavailability**: If `push_to_reports_queue` failed (e.g. RabbitMQ down), errors were only logged and `Ok(())` was still returned, so the checkpoint moved forward and the failed triggers were never retried.
3. **Silent cache errors**: The result of `cache.insert(...)` was discarded with `let _ =`, hiding any cache write failures.

## Fix

- **Incremental checkpoint advancement**: The checkpoint is now updated per hour bucket, only after all reports in that bucket have been successfully enqueued.
- **Fail-stop on enqueue errors**: If any report fails to enqueue, checkpoint advancement stops immediately. The failed hour bucket (and any remaining ones) will be retried on the next scheduler tick.
- **Error propagation for cache writes**: `cache.insert(...)` errors are now propagated via `?` instead of silently ignored, so the caller can log and handle them.
- **Empty-window handling**: When no hour boundaries need processing, the checkpoint is still advanced to `now` to prevent re-scanning the same sub-hour window.

## Testing

- `cargo check` passes with no compilation errors.
- The change is minimal and surgical — only `app-server/src/reports/scheduler.rs` is modified (23 insertions, 4 deletions).
- Behavioral correctness follows from the invariant that the checkpoint can never advance past a timestamp whose reports have not all been successfully enqueued.

---

> [!NOTE]
> **Medium Risk**
> Changes the report scheduling checkpointing semantics and error handling; a mistake could cause duplicate triggers or stalled scheduling if failures aren’t handled as expected.
> 
> **Overview**
> Fixes the reports scheduler so the `REPORT_SCHEDULER_LAST_CHECK_CACHE_KEY` checkpoint is **not** advanced until report triggers are actually enqueued.
> 
> Checkpoint writes now happen per processed hour bucket (and propagate cache write errors), while enqueue failures cause the function to error out and **hold the checkpoint** for retry; when there are no hour boundaries, the checkpoint still advances to `now` to avoid rescanning the same sub-hour window.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3279eb56ccacac82cdc88e85eea2bb011be72b46. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/app-server/src/reports/scheduler.rs
+++ b/app-server/src/reports/scheduler.rs
@@ -96,12 +96,13 @@ async fn check_and_enqueue(
         now
     );
 
-    let _ = cache
-        .insert(REPORT_SCHEDULER_LAST_CHECK_CACHE_KEY, now.timestamp())
-        .await;
-
     let hours_to_check = hour_boundaries_between(last_check, now);
     if hours_to_check.is_empty() {
+        // No hour boundaries to process; advance checkpoint so the same
+        // sub-hour window is not re-scanned on the next tick.
+        cache
+            .insert(REPORT_SCHEDULER_LAST_CHECK_CACHE_KEY, now.timestamp())
+            .await?;
         return Ok(());
     }
     log::info!(
@@ -113,6 +114,7 @@ async fn check_and_enqueue(
     for (weekday, hour, triggered_at) in hours_to_check {
         let reports = get_reports_for_weekday_and_hour(pool, weekday, hour).await?;
 
+        let mut all_enqueued = true;
         for report in reports {
             let message = ReportTriggerMessage {
                 id: report.id,
@@ -129,8 +131,23 @@ async fn check_and_enqueue(
                     report.id,
                     e
                 );
+                all_enqueued = false;
             }
         }
+
+        if !all_enqueued {
+            // Stop advancing the checkpoint so this hour bucket (and any
+            // remaining ones) will be retried on the next tick.
+            return Err(anyhow::anyhow!(
+                "One or more reports failed to enqueue for hour {triggered_at}; checkpoint held for retry"
+            ));
+        }
+
+        // Advance checkpoint only after all reports in this hour bucket
+        // have been successfully enqueued.
+        cache
+            .insert(REPORT_SCHEDULER_LAST_CHECK_CACHE_KEY, triggered_at)
+            .await?;
     }
 
     Ok(())


ixchio:
Hey @dinmukhamedm  this addresses #1586, the checkpoint race in the reports scheduler. The core issue is that the checkpoint was moving forward before the enqueue loop finished, so a crash or MQ hiccup would silently drop scheduled reports with no retry.

The fix is pretty small — just defers the checkpoint write to after each hour bucket is fully enqueued, and returns an error on partial failure so it actually shows up in logs. I also addressed the Greptile bot's review feedback (returning Err instead of Ok on failure for proper error-level logging).

Re: the duplicate notification concern the bot flagged — that's a pre-existing at-least-once delivery gap in the consumer, not something this PR introduces. Happy to open a follow-up issue for adding idempotency on the generator side if that'd be useful.

Would appreciate a review when you get a chance, thanks!

ixchio:
closing it lack of reponse

PR #1630, opened 2026-04-12
fix: handle event proper escaping and not retriable

Rainhunter13:
> [!NOTE]
> **Medium Risk**
> Touches signal run retry behavior and event creation flow by changing what gets retried and adding transient backoff retries around ClickHouse inserts/notification processing; misclassification could increase permanent failures or duplicate work. Also changes JSON attribute rewriting logic, which could affect event attribute formatting/linking.
> 
> **Overview**
> **Signal run retry semantics are simplified and tightened.** `FailureMetadata` now stores a single `retryable` flag, and `retry_or_fail_runs` uses it directly; event-creation failures explicitly mark runs as *non-retryable*.
> 
> **Event creation is made more resilient but time-bounded.** `handle_create_event` now wraps both `insert_signal_events` and `process_event_notifications_and_clustering` in exponential backoff retries (up to ~10s) before failing.
> 
> **Span reference rewriting is safer for JSON and tricky strings.** `replace_span_tags_with_links` now walks the JSON value tree and rewrites only string fields (instead of stringifying/parsing the whole JSON), with shared compiled regexes and an added test covering `<span ... name='..."..."' />` cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a715d65adef2325f8bf3fbbd09da8678900de15c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/app-server/src/signals/response_processor.rs
+++ b/app-server/src/signals/response_processor.rs
@@ -1,8 +1,10 @@
+use backoff::ExponentialBackoffBuilder;
 use regex::Regex;
 use serde::Serialize;
 use std::{
     collections::{HashMap, HashSet},
     sync::{Arc, LazyLock},
+    time::Duration,
 };
 use uuid::Uuid;
 
@@ -55,13 +57,14 @@ pub enum StepResult {
     Failed {
         error: String,
         finish_reason: Option<FinishReason>,
+        // Whether the error is due to processing the response, not the provider.
+        // Usually due to malformed function calls so is retryable.
         is_processing_error: bool,
     },
 }
 
 pub struct FailureMetadata {
-    pub finish_reason: Option<FinishReason>,
-    pub is_processing_error: bool,
+    pub retryable: bool,
 }
 
 pub struct ProcessedResponses {
@@ -147,6 +150,7 @@ pub async fn process_provider_responses(
         .await;
 
         new_messages.extend(new_run_messages);
+        log::error!("STEP RESULT: {:?}", step_result);
 
         match step_result {
             StepResult::CompletedNoEvent => {
@@ -176,6 +180,7 @@ pub async fn process_provider_responses(
                     }
                     Err(e) => {
                         log::error!("[SIGNAL JOB] Failed to create event: {:?}", e);
+                        failure_metadata.insert(run.run_id, FailureMetadata { retryable: false });
                         failed_runs.push(run.failed(format!("Failed to create event: {}", e)));
                     }
                 }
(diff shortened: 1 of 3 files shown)


Rainhunter13:
@laminar-coding-agent /autoreply start

app-server/src/signals/response_processor.rs
```diff
@@ -776,15 +783,54 @@ pub async fn handle_create_event(
...
+    let backoff = ExponentialBackoffBuilder::new()
+        .with_initial_interval(Duration::from_secs(1))
+        .with_multiplier(2.0)
+        .with_max_elapsed_time(Some(Duration::from_secs(10)))
+        .build();
+    let ch_ref = &clickhouse;
+    let event_ref = &signal_event;
+    // NOTE: all errors treated as transient because anyhow::Error doesn't expose
+    // ClickHouse error kinds; permanent errors (schema mismatch, etc.) will retry
+    // until the 10s timeout before propagating.
```
dinmukhamedm, line 796: well ideally we should expose the error up to here, or maybe more broadly create an error enum transient / permanent both containing anyhow, similar to what we have in message handlers. But ok for now

app-server/src/signals/utils.rs
```diff
@@ -222,13 +232,19 @@ pub fn replace_span_tags_with_links(
...
+fn replace_span_tags_in_str(
+    s: &str,
+    span_ids_map: &HashMap<String, Uuid>,
+    project_id: Uuid,
+    trace_id: Uuid,
+) -> String {
+    // 1. Replace proper <span id='...' name='...' /> XML tags.
+    //    The LLM prompt in prompts.rs explicitly instructs this attribute order (id before name),
+    //    so we rely on it here rather than making the pattern order-agnostic.
+    let after_xml = SPAN_XML_TAG_RE.replace_all(s, |caps: &regex::Captures| {
```
dinmukhamedm, line 247: Good call, and probably out of scope of this PR (was like this anw), but I don't think an instruction in prompts is strict enough to enforce id before name. And XML attributes order does not matter (generally for XML format), so I think we should make the pattern order agnostic. Or simpler, two patterns

dinmukhamedm (approved):

app-server/src/signals/response_processor.rs
Rainhunter13: agreed

PR #1820, opened 2026-05-28
log error on invalid llm default headers

Rainhunter13:
> [!NOTE]
> **Low Risk**
> Observability-only change; configuration gating behavior is unchanged aside from a single warn log.
> 
> **Overview**
> When **`LLM_DEFAULT_HEADERS_JSON`** fails validation, **`hasValidLlmDefaultHeaders()`** still treats the AI provider as unconfigured (same behavior for **`isAiProviderConfigured()`**), but it now emits a **one-time** **`console.warn`** with the parse/validation error instead of failing silently.
> 
> A module-level flag prevents repeated warnings on every check.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2ca6fc5fdf74142dce1e56e74b2b39fc6b25d07a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/frontend/lib/ai/model.ts
+++ b/frontend/lib/ai/model.ts
@@ -92,7 +92,9 @@ function hasValidLlmDefaultHeaders(): boolean {
   try {
     parseLlmDefaultHeaders();
     return true;
-  } catch {
+  } catch (error) {
+    const message = error instanceof Error ? error.message : String(error);
+    console.warn(`AI provider failed to configure: ${message}`);
     return false;
   }
 }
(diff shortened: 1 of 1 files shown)


dinmukhamedm (approved):
@Rainhunter13 what was the rationale here? We can merge, but generally I think eslint punishes console.log statements, and by default console.warn is no different from console.log.

As a followup effort we could add a pretty logger to our next deployment, e.g. pino

Rainhunter13:
> @Rainhunter13 what was the rationale here? We can merge, but generally I think eslint punishes console.log statements, and by default console.warn is no different from console.log.
> 
> As a followup effort we could add a pretty logger to our next deployment, e.g. pino

Ah yeah small cleanup so self hosted users will see an error in logs in case of wrong headers, rather than silently llm client not working. Hm don't we use console log normally?

dinmukhamedm:
> Hm don't we use console log normally?

We do, but linter punishes that and for a good reason. It's fine for now. I created a super low prio ticket for this (1800)

PR #1856, opened 2026-06-04
feat(oauth): CLI OAuth 2.0 Device Authorization Grant

kolbeyang:
## Summary
- New OAuth 2.0 Device Authorization Grant (RFC 8628) on the Next.js frontend so the lmnr CLI can authenticate without users pasting an API key.
- Rust app-server gains a JWT bearer auth path that validates RS256 tokens via JWKS fetched from the frontend and synthesizes the same `ProjectApiKey` request extension that the existing API-key middleware produces. All existing protected routes accept either credential with no per-route changes.
- Companion CLI PR at lmnr-ai/lmnr-ts adds `login`/`logout`/`status` commands and retrofits `dataset`/`sql`/`dev` to use stored tokens with auto-refresh.

## What's in this PR

**Frontend (OAuth provider)**
- 3 new endpoints: `POST /oauth/device/authorize`, `POST /oauth/token` (device-code + refresh-token grants), `GET /oauth/jwks`, plus `GET /.well-known/openid-configuration` for discovery.
- Server-rendered approval page at `/oauth/device` with project picker. Signed-in users only; deep link preserved through sign-in via a new `x-pathname` header forwarded by `proxy.ts` into the `(auth)` layout.
- DB migration `0087_oauth_device_flow.sql` adds three tables: `oauth_signing_keys` (RS256 keypair, AEAD-wrapped private key), `oauth_device_codes`, `oauth_refresh_tokens` (SHA3-256 hashed, single-use rotation with `family_id` reuse detection).
- Signing-key boot hook in `frontend/instrumentation.ts` mints the active RS256 keypair on first run.
- `/api/cli/me` accepts both session cookies (web) and JWT bearer (CLI post-login enrichment).

**Rust (JWT validation)**
- `app-server/src/auth/jwks.rs` — in-process JWKS cache (10-min fresh TTL, 24-h stale fallback, force-refresh on unknown `kid` throttled).
- `app-server/src/auth/jwt.rs` — RS256 validator with iss/aud checks. Synthesizes `ProjectApiKey { hash: "jwt:<jti>", ... }`.
- `project_validator_with_jwt` middleware dispatcher routes bearer tokens that look like JWTs to the new path; everything else falls through to the existing API-key validator. Wired in `main.rs` and `traces/grpc_service.rs` (HTTP + gRPC ingestion both honor JWT bearer).
- New dep: `jsonwebtoken = "9"`.

**Tests**
- Frontend: signing-key roundtrip + JWT verify (2 cases).
[...]

PR #1882, opened 2026-06-09
feat(cli-better-auth): CLI Access Token Auth

kolbeyang:
Supersedes #1856 (the pre-BetterAuth, hand-rolled device-flow approach).

Built on the BetterAuth migration (base: `feat/lam-1720-better-auth`).

## App-server
- New `/v1/cli/*` surface authenticated by the BetterAuth **user JWT**, parallel to the unchanged `/v1/*` project-API-key surface. `ProjectContext` unified principal; `sql` / `datasets` / `rollouts` twinned. Project comes from the `x-lmnr-project-id` header; workspace membership is checked per request (JWKS verify, 60s membership cache).
- `GET /v1/cli/projects` (JWT-only) for CLI project discovery.

## Frontend
- BetterAuth device-flow `/device` verification page (device code shown as a subtle full-width box).
- `device_codes` migration (0090).

**Config:** the app-server must set `NEXT_PUBLIC_URL` (the JWKS source `${NEXT_PUBLIC_URL}/api/auth/jwks`).

---

> [!NOTE]
> **High Risk**
> New authentication and authorization on app-server (JWKS, membership cache, parallel API surface) plus device-flow metadata ordering; mistakes could allow cross-project access or break CLI login.
> 
> **Overview**
> Adds a **BetterAuth user-JWT** path on app-server at **`/v1/cli/*`**, alongside the existing project API key **`/v1/*`** surface. Bearer tokens are verified against the frontend JWKS (`CliUserAuth`); project-scoped routes require **`x-lmnr-project-id`** and a cached workspace membership check (`CliProjectAuth`). New discovery **`GET /v1/cli/projects`**; twins for SQL, datasets, trace metadata, and debugger session rename. Dataset create/list logic is extracted into **`datasets::service`** so API-key and CLI handlers share behavior; rollout **`PATCH …/name`** moves from the SDK rollout routes to CLI-only. SQL per-project rate limiting runs inside shared **`handle_sql_query`** after auth resolves `project_id`. **`GET /v1/project`** returns the project behind an API key.
> 
> On the frontend, **`/device`** drives CLI login: verify code, pick or create a project, then approve via **`/api/cli/device/approve`** (metadata written before approve). **`POST /api/auth/.../device/token`** is wrapped to attach **`x-lmnr-metadata`** for the polling CLI; **`POST /api/cli/api-key`** mints project keys for **`lmnr-cli setup`**. Sign-in/sign-up UI is consolidated around **`AuthProviderButtons`** with landing-style styling. **`CLAUDE.md`** documents CLI device flow and auth edge cases in more detail.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e0add3d1bad46f8a38e2585defdffdb684e3db1d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


app-server/src/api/v1/cli.rs
```diff
@@ -0,0 +1,27 @@
+use actix_web::{HttpResponse, get, web};
+use serde_json::json;
+
+use crate::auth::cli_user::CliUser;
+use crate::db::project_api_keys::ProjectApiKey;
+use crate::db::{self, DB};
+
+/// `GET /v1/cli/projects` — workspaces + projects the authenticated user can
```
kolbeyang: This comment is too long, one ine please

app-server/src/api/v1/cli.rs
```diff
@@ -0,0 +1,27 @@
...
+/// Lets the CLI discover and select a project after `login`.
+#[get("")]
+pub async fn list_projects(user: CliUser, db: web::Data<DB>) -> actix_web::Result<HttpResponse> {
+    let projects = db::projects::get_projects_for_user(&db.pool, &user.user_id)
+        .await
+        .map_err(actix_web::error::ErrorInternalServerError)?;
+    Ok(HttpResponse::Ok().json(json!({ "projects": projects })))
+}
+
+/// `GET /v1/cli/whoami` — returns the project id behind the supplied project API
```
kolbeyang: This comment is too long, one line please

app-server/src/auth/mod.rs
```diff
@@ -65,7 +101,15 @@ async fn validate_project_api_key(
                     req,
                 ));
             }
+            // Additive: also insert the unified ProjectContext so the shared
```
kolbeyang: Remove comment

app-server/src/auth/mod.rs
```diff
@@ -29,6 +32,39 @@ impl FromRequest for ProjectApiKey {
...
+pub enum AuthPrincipal {
+    ApiKey(ProjectApiKey),
+    User { user_id: Uuid },
+}
+
+/// Unified principal consumed by the shared handlers (sql/datasets/rollouts).
+/// `project_id` is resolved by the auth middleware — from the API key for the
+/// key path, or from the `x-lmnr-project-id` header for the CLI user path.
+#[derive(Clone)]
+pub struct ProjectContext {
```
kolbeyang: Rename to ProjectAuthContext for clarity

Then rename principal to credential which is an enum of either ApiKey(ProjectApiKey) or UserToken { user_id }

app-server/src/auth/cli_user.rs
kolbeyang, line 1: Hmm let's try to use a library to reduce some of the boilerplate here
  jwks_client_rs (prima) — the right pick
  - v0.6.0, published May 2026, ~14k downloads/mo, MIT, actively maintained. ~940 LOC,
  lightweight.
  - Built on top of jsonwebtoken — the same crate you already use — plus reqwest. Internal
  cache + automatic timed refresh, which is exactly your hand-rolled JwksCache.
  - Configurable TLS (rustls / aws-lc / native-tls), so no forced OpenSSL — matches your
  rustls setup.
  - EdDSA caveat: not documented explicitly, but it passes Algorithm straight through to
  jsonwebtoken, which supports EdDSA (that's what your code uses today). Verify with one
  of your existing unit tests after swapping.
  - Version caveat (the real one): v0.6 pins jsonwebtoken 10.2 + reqwest 0.13, but you're
  on jsonwebtoken 9.3 + reqwest 0.12. Adopting it adds a duplicate jsonwebtoken 9↔10 in
  the tree (reqwest 0.13 is already transitively present). Either accept the dup, or pin
  an older jwks_client_rs (0.5.x) that targets jsonwebtoken 9, or bump your own
  jsonwebtoken to 10.

Let's dump jsonwebtoken -> 10 and reqwest -> 0.13 (double check migration guides for reqwest please to make sure no breaking changes)

app-server/src/db/projects.rs
```diff
@@ -169,6 +169,62 @@ pub async fn get_projects_for_workspace(
     Ok(projects)
 }
 
+/// Returns true if `user_id` is a member of the workspace that owns `project_id`.
```
kolbeyang, line 172: Reduce comment to one line

kolbeyang (commented):
More on the way, publishing now so Claude can get started

app-server/src/main.rs
```diff
@@ -1834,6 +1845,39 @@ fn main() -> anyhow::Result<()> {
                                     .wrap(project_auth.clone())
                                     .service(api::v1::sql::execute_sql_query),
                             )
+                            // User-scoped CLI discovery (no project header) — must
```
kolbeyang: Let's rename these please

cli_user_jwt_auth -> cli_user_auth (only authenticates user right?)
cli_user_auth -> cli_project_auth (verifies user has access to a project right?)

I know we pass ProjectApiKey to every handler or whatever, but let's rely on ProjectAuthContext for the project_id instead. Remove ProjectApiKey from all the handlers (except via the ProjectAuthContext).

This way the rate limiter can key off of the ProjectAuthContext instead of off of the ProjectApiKey, 

ProjectAuthContext should be the single unified source of truth for authentication state.

kolbeyang (commented):
More comments coming

frontend/lib/actions/cli-auth/index.ts
```diff
@@ -0,0 +1,127 @@
...
+  if (!session?.user) return { error: "Unauthorized" };
+
+  const userCode = normalizeUserCode(rawUserCode);
+  if (userCode.length === 0) return { error: "Invalid code" };
+
+  const context = await loadDeviceContext(userCode);
+  if (!context) return { error: "We couldn't find that code." };
+  if (context.status !== "pending") return { error: "This code has already been processed." };
+
+  // 1) Write the chosen project into `scope` WHILE the row is pending. deviceApprove
```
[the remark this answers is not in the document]
kolbeyang: Will investigate but project-id in scope doesn't actually give access

app-server/src/api/v1/browser_sessions.rs
```diff
@@ -64,7 +65,7 @@ pub struct EventBatch {
 #[post("events")]
 async fn create_session_event(
     batch: web::Json<EventBatch>,
-    project_api_key: ProjectApiKey,
+    ctx: ProjectAuthContext,
```
kolbeyang: I don't like ctx because it's a generic name, let's rename everywhere to project_auth_ctx, it's more clear

app-server/src/auth/cli_user.rs
```diff
@@ -0,0 +1,556 @@
...
+    /// retained for call-site compatibility; `WebSource` manages its own client.
+    pub fn from_env(_http: reqwest::Client) -> Self {
+        let base = match std::env::var("NEXT_PUBLIC_URL") {
+            Ok(v) => v.trim_end_matches('/').to_string(),
+            Err(_) => {
+                log::warn!(
+                    "NEXT_PUBLIC_URL unset; CLI-auth JWKS will default to https://lmnr.ai"
+                );
+                "https://lmnr.ai".to_string()
+            }
```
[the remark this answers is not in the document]
kolbeyang: App server uses NEXT_PUBLIC_URL by design, BETTER_AUTH_URL does not need to be the same as NEXT_PUBLIC_URL

NEXT_PUBLIC_URL is used by app-server to fetch the jwks

app-server/src/api/v1/cli/mod.rs
```diff
@@ -0,0 +1,29 @@
...
+pub mod sql;
+pub mod traces;
+
+use actix_web::{HttpResponse, get, web};
+use serde_json::json;
+
+use crate::auth::cli_user::CliUserAuth;
+use crate::db::{self, DB};
+
+/// `GET /v1/cli/projects` — user-scoped project discovery. Takes `CliUserAuth`
```
kolbeyang: Reduce comment to 1 line

app-server/src/api/v1/cli/rollouts.rs
```diff
@@ -0,0 +1,57 @@
...
+    routes::types::ResponseResult,
+};
+
+#[derive(serde::Deserialize)]
+#[serde(rename_all = "camelCase")]
+struct UpdateNameRequest {
+    name: String,
+}
+
+/// `PATCH /v1/cli/rollouts/{session_id}/name` — rename a debug session.
```
kolbeyang: Reduce comment to either 1 or 2 lines

app-server/src/api/v1/cli/sql.rs
```diff
@@ -0,0 +1,59 @@
...
+    api::v1::sql::{SqlQueryRequest, run_sql_query},
+    auth::cli_user::CliProjectAuth,
+    cache::Cache,
+    db::DB,
+    query_engine::QueryEngine,
+    routes::types::ResponseResult,
+    sql::ClickhouseReadonlyClient,
+};
+
+/// `POST /v1/cli/sql/query` — CLI twin of `/v1/sql/query`. User-token auth via
```
kolbeyang: Reduce comment to max 4 lines

app-server/src/main.rs
```diff
@@ -1796,11 +1797,22 @@ fn main() -> anyhow::Result<()> {
...
+                    let jwks_cache = web::Data::new(Arc::new(
+                        auth::cli_user::JwksCache::from_env(http_client_for_http.clone()),
+                    ));
+
                     log::info!("Spinning up full HTTP server");
                     HttpServer::new(move || {
                         let project_auth = HttpAuthentication::bearer(auth::project_validator);
                         let project_ingestion_auth =
                             HttpAuthentication::bearer(auth::project_ingestion_validator);
+                        // Single authN middleware for the whole /v1/cli scope:
```
kolbeyang: Reduce to max 2 line comment

app-server/src/main.rs
```diff
@@ -1898,19 +1911,48 @@ fn main() -> anyhow::Result<()> {
                                     .wrap(project_auth.clone())
                                     .service(api::v1::sql::execute_sql_query),
                             )
+                            // CLI user-token surface. ONE scope, ONE authN
```
kolbeyang: Reduce comment to max 2 lines please, you cited implementation details from our discussion not documentation that's useful in the future

app-server/Cargo.toml
```diff
@@ -49,7 +51,11 @@ rand = "0.9.4"
 rayon = "1.11.0"
 redis = {version = "0.32.7", features = ["tokio-comp"]}
 regex = "1.12.3"
-reqwest = {version = "0.12.22", features = ["json", "stream"]}
+# reqwest 0.13 made rustls the default TLS backend (was native-tls in 0.12) and its default `rustls`
```
kolbeyang: Remove comment please

app-server/src/datasets/service.rs
```diff
@@ -0,0 +1,254 @@
...
+    http_client: Arc<reqwest::Client>,
+    db: Arc<DB>,
+    cache: Arc<Cache>,
+) -> anyhow::Result<Option<(Vec<Datapoint>, u64)>> {
+    let dataset_id = match dataset {
+        DatasetIdentifier::Name(name) => {
+            match db::datasets::get_dataset_id_by_name(&db.pool, &name.dataset_name, project_id)
+                .await?
+            {
+                Some(id) => id,
```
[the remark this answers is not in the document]
kolbeyang: Fixed in cce39d9d5 — the `Id` branch now calls `db::datasets::dataset_exists(&db.pool, id.dataset_id, project_id)` (same gate the write path uses) and returns `None` → 404 for a foreign/absent id, so cross-project reads are blocked without leaking dataset existence.

app-server/src/api/v1/projects.rs
dinmukhamedm, line 1: [not actionable, just thinking out loud] one other way we could do this is to list all project API key previews (called shorthand in DB) given a project ID, and match on the CLI side. Not sure which one is more secure (both seem kinda insecure), so let's do what fits the CLI flow better.

app-server/src/api/v1/cli/sql.rs
```diff
@@ -0,0 +1,74 @@
...
+    query_engine::QueryEngine,
+    routes::types::ResponseResult,
+    sql::{self, ClickhouseReadonlyClient},
+};
+
+/// `POST /v1/cli/sql/query` — CLI twin of `/v1/sql/query`. Rate limiting is
+/// inline (not scope middleware) because the extractor resolves project_id after
+/// middleware; shares the `ratelimit:<project_id>` key with `/v1/sql`. Fail-open.
+#[post("query")]
+pub async fn execute_sql_query(
```
dinmukhamedm, line 20: nit: was the decision to copy and paste code instead of a shared function based on the fact that code is super thin?

Just thinking in terms of maintainability, what if we want to change something (and we may given that we are about to migrate query engine to Rust), we need to keep in mind updating two places.

app-server/src/db/projects.rs
```diff
@@ -169,6 +169,59 @@ pub async fn get_projects_for_workspace(
...
+/// All projects in workspaces the user is a member of (any role), ordered by
+/// workspace then project name. Used by the CLI to let the user discover and
+/// select a project after `login`.
+pub async fn get_projects_for_user(pool: &PgPool, user_id: &Uuid) -> anyhow::Result<Vec<CliProject>> {
+    let projects = sqlx::query_as::<_, CliProject>(
+        "SELECT p.id, p.name, w.id AS workspace_id, w.name AS workspace_name
+         FROM projects p
+         JOIN members_of_workspaces m ON m.workspace_id = p.workspace_id
+         JOIN workspaces w ON w.id = p.workspace_id
+         WHERE m.user_id = $1
```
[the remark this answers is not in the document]
dinmukhamedm: unlikely to be even tens

app-server/Cargo.toml
```diff
@@ -49,7 +51,7 @@ rand = "0.9.4"
 rayon = "1.11.0"
 redis = {version = "0.32.7", features = ["tokio-comp"]}
 regex = "1.12.3"
-reqwest = {version = "0.12.22", features = ["json", "stream"]}
+reqwest = {version = "0.13", default-features = false, features = ["json", "stream", "rustls-no-provider", "charset", "http2", "system-proxy"]}
```
dinmukhamedm: is this really needed in this PR? Reading the number of breaking changes https://github.com/seanmonstar/reqwest/blob/master/CHANGELOG.md#v0130 and the amount of features we had to change, I'd ideally prefer not to upgrade the library. We should do that separately.

I'm most concerned about unrelated places where we use request client directly (not transitively via LLM SDKs, as they probably pin to a certain version), e.g. some of signal providers.

frontend/lib/db/migrations/0091_lam_1720_device_code.sql
dinmukhamedm, line 1: FYI, I have applied these to Prod DB

frontend/instrumentation.ts
```diff
@@ -294,6 +294,23 @@ export async function register() {
...
       console.log("Local DB is not enabled, skipping migrations and initial data");
     }
+
+    // Prune expired device-flow codes once per hour — runs in prod AND local.
+    // BetterAuth's plugin only deletes rows opportunistically (on poll / approve
+    // / deny); a user who walks away leaves the row indefinitely. Keep ~24 hours
+    // of history so the verification page can still surface "expired" banners
+    // instead of "unknown code". Dynamic-import db here since the LOCAL_DB block
+    // above scopes its own import.
+    const { db: prodDb } = await import("@/lib/db/drizzle.ts");
```
dinmukhamedm: what's the intent here? Keep the prod DB (i.e. Laminar Cloud) up to date? If so, this can live as a cron job in pg_cron on our Prod DBs, not here

app-server/src/auth/cli_user.rs
```diff
@@ -0,0 +1,519 @@
...
+    client: JwksClient<WebSource>,
+}
+
+impl JwksCache {
+    /// Build from `NEXT_PUBLIC_URL` (same var notifications use), deriving the
+    /// JWKS endpoint `{base}/api/auth/jwks`. Falls back to `https://lmnr.ai`
+    /// with a warning so self-hosters notice an unset var. The `http` arg is
+    /// retained for call-site compatibility; `WebSource` manages its own client.
+    pub fn from_env(_http: reqwest::Client) -> Self {
+        let base = match std::env::var("NEXT_PUBLIC_URL") {
```
dinmukhamedm: Hmmm, I think this could (should?) be a separate env var so that we are able to call the internal frontend service instead of the public URL. So probably let's do something like

NEXT_INTERNAL_URL -> NEXT_PUBLIC_URL -> laminar.sh

One thing to verify though is whether this flow requires HTTPs. This would be somewhat silly, but then we have to go through public URL.

If we do introduce this variable, let's not forget to set it on our k8s deployment prior to the release

app-server/src/auth/cli_user.rs
```diff
@@ -0,0 +1,519 @@
...
+        Self::from_url(&jwks_url)
+    }
+
+    /// Build a client pointing at an explicit JWKS URL with the standard TTL.
+    fn from_url(jwks_url: &str) -> Self {
+        // A malformed URL here is a misconfiguration; fall back to the default
+        // so the process still boots (verification just won't resolve keys).
+        let url = Url::parse(jwks_url).unwrap_or_else(|e| {
+            log::error!("invalid JWKS URL {jwks_url:?}: {e}; CLI auth will not resolve keys");
+            Url::parse("https://lmnr.ai/api/auth/jwks").expect("static fallback URL parses")
```
dinmukhamedm: same here, laminar.sh, not lmnr.ai

app-server/src/auth/cli_user.rs
```diff
@@ -0,0 +1,519 @@
...
+    project_id: Uuid,
+) -> anyhow::Result<bool> {
+    let cache_key = format!("{PROJECT_MEMBERSHIP_CACHE_KEY}:{user_id}:{project_id}");
+    if let Ok(Some(is_member)) = cache.get::<bool>(&cache_key).await {
+        return Ok(is_member);
+    }
+    let is_member = crate::db::projects::project_has_member(pool, &user_id, &project_id).await?;
+    let _ = cache
+        .insert_with_ttl::<bool>(&cache_key, is_member, PROJECT_MEMBERSHIP_TTL_SECONDS)
+        .await;
```
dinmukhamedm: nit: log warn / error on error?

app-server/src/api/v1/browser_sessions.rs
dinmukhamedm, line 1: super nit: we should probably restore the changes on this and other similar files

dinmukhamedm (changes requested):

dinmukhamedm:
@olzhik11 can you please review the `frontend/` changes?

dinmukhamedm (approved):

olzhik11 (approved):

PR #1888, opened 2026-06-11
refactor(debugger): DX follow-up

kolbeyang:
…ring project-key route

Extract a shared handle_register_session in api/v1/rollouts.rs and have both the project-API-key POST /v1/rollouts/{id} and the new CLI user-token POST /v1/cli/rollouts/{id} (CliProjectAuth) delegate to it. Register the CLI route in the /v1/cli scope. Project-key behavior is unchanged.

---

> [!NOTE]
> **Medium Risk**
> Touches realtime session lifecycle, concurrent span merge semantics, and new app-server routes for register/rename; incorrect merge or auth wiring could cause stale UI or failed CLI registration, but changes are scoped to debugger/session views rather than core ingestion.
> 
> **Overview**
> **Debugger session naming** gets end-to-end support: **CLI** `POST /v1/cli/rollouts/{session_id}` mirrors the SDK register upsert with user-token auth; the **UI** renames via a Next `PATCH` that calls app-server `PATCH /api/v1/projects/{projectId}/rollouts/{id}/name`, which persists and broadcasts **`session_update`** (matching the CLI rename path). The session view adds an **editable title** (`sessionNameRaw` vs breadcrumb fallback) and handles **`session_deleted`** with a toast and redirect.
> 
> **Debugger UX and data loading** add a **per-trace condensed timeline** in the debugger session panel (toggle on trace cards, shared session-store zoom/selection state, refactored timeline leaves to accept props instead of trace-view store). Span loading is reworked: **`ensureTraceSpans` → `fetchTraceSpans`** (debugger override always refetches on expand with deduped in-flight state), realtime span buffering is removed in favor of unconditional upserts, and hydrate/expand merges prefer fresher SSE spans over lagging CH snapshots. **Stick-to-bottom** waits until initial traces load; empty sessions show a message; note **span chips** can resolve type via new **`GET .../spans/{spanId}/type`**.
> 
> **Supporting pieces:** `NotFoundError` for 404 mapping on rename; narrow layout tweaks (`min-w`, overflow); scroll-to-subagent-group from the condensed timeline; trace expand no longer blocks on a pending spinner before opening.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit de79bd1c81f6faf91eb63272f18589f59e7971e9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/app-server/src/api/v1/rollouts.rs
+++ b/app-server/src/api/v1/rollouts.rs
@@ -33,10 +33,26 @@ pub async fn register_session(
     body: web::Json<RegisterSessionRequest>,
     db: web::Data<DB>,
 ) -> ResponseResult {
+    handle_register_session(
+        path.into_inner(),
+        project_api_key.project_id,
+        body.into_inner().name,
+        db,
+    )
+    .await
+}
+
+/// Shared register (idempotent upsert) handler used by both the project-API-key
+/// `POST /v1/rollouts/{session_id}` and the CLI user-token
+/// `POST /v1/cli/rollouts/{session_id}` routes. Differs only in how the
+/// `project_id` is authenticated; the DB call and JSON response are identical.
+pub async fn handle_register_session(
+    session_id: Uuid,
+    project_id: Uuid,
+    name: Option<String>,
+    db: web::Data<DB>,
+) -> ResponseResult {
     let db = db.into_inner();
-    let session_id = path.into_inner();
-    let project_id = project_api_key.project_id;
-    let name = body.into_inner().name;
 
     let session =
         create_or_update_debugger_session(&db.pool, &session_id, &project_id, name).await?;
--- a/app-server/src/api/v1/rollouts.rs
+++ b/app-server/src/api/v1/rollouts.rs
@@ -33,26 +33,10 @@ pub async fn register_session(
     body: web::Json<RegisterSessionRequest>,
     db: web::Data<DB>,
 ) -> ResponseResult {
-    handle_register_session(
-        path.into_inner(),
-        project_api_key.project_id,
-        body.into_inner().name,
-        db,
-    )
-    .await
-}
-
-/// Shared register (idempotent upsert) handler used by both the project-API-key
-/// `POST /v1/rollouts/{session_id}` and the CLI user-token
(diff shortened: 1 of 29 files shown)


app-server/src/api/v1/rollouts.rs
```diff
@@ -32,11 +32,27 @@ pub async fn register_session(
...
+        db,
+    )
+    .await
+}
+
+/// Shared register (idempotent upsert) handler used by both the project-API-key
+/// `POST /v1/rollouts/{session_id}` and the CLI user-token
+/// `POST /v1/cli/rollouts/{session_id}` routes. Differs only in how the
+/// `project_id` is authenticated; the DB call and JSON response are identical.
+pub async fn handle_register_session(
```
kolbeyang: Do we realy need this helper, looking at it is this not nearly identical to the create_or_update_debugger_session? unless I'm missing something removing this utliity doesn't actually hurt too much on code duplication.

kolbeyang: Good call — removed the `handle_register_session` wrapper in 9b94bda80. It only dedup'd the `create_or_update_debugger_session` call + JSON response, so I inlined those ~2 lines into both handlers (each sourcing `project_id` from its own auth extractor). `RegisterSessionRequest` stays shared. `cargo check` clean.

dinmukhamedm (commented):
Logic looks sound to me, some of the cursor bugbot comments are valid. The rest is frontend, @olzhik11 can you please take a look?

frontend/components/debugger-sessions/debugger-session-view/span-reference.tsx
```diff
@@ -1,11 +1,72 @@
...
+      const data = (await res.json()) as { spanType?: SpanType };
+      return data.spanType ?? null;
+    } catch {
+      // Don't poison the cache — a later mount can retry.
+      spanTypeCache.delete(key);
+      return null;
+    }
+  })();
+  spanTypeCache.set(key, promise);
+  return promise;
```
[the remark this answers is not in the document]
kolbeyang: Span ids are unique

PR #2153, opened 2026-07-28
feat: carry agent_input live on realtime trace-update payload

kolbeyang:
## What

Makes the extracted trace **input** (`agent_input`, `Feature::InputExtraction`) a first-class field on the realtime `trace_update` SSE payload, so the debugger session view (and the traces table) show the input **live** without a manual reload.

## Why

`agent_input` is written **asynchronously** — an LLM regex round-trip that finishes seconds after the trace lands (output is an instant inline heuristic, so it already appears immediately). The debugger loads each trace row **once** and never refetches; its `trace_update` handler only patched `hasBrowserSession`, so the input pill stayed empty until reload.

The regular trace view looks instant only because opening it is a fresh fetch by which time extraction has finished — it doesn't live-update `agent_input` either.

Regression origin: PR #2120 / LAM-2006 switched the debugger's collapsed-body input source from a span-derived read-time preview to the stored `agent_input` column. The DB value is correct; it just never reaches the already-rendered row live.

## Approach — Option B (thread the value)

Rather than reading `agent_input` off the deprecated `lmnr_user_task` metadata fold (Option A, on a demolition schedule), the extracted value is threaded straight from the ingestion path (`raw_trace_io`) to the realtime dispatch as a first-class field on **both** `RealtimeTrace` (traces table) and `RealtimeDebuggerTrace` (debugger). It's JSON-stringified (`value.to_string()`) to match `CHTraceAgentInput.value` / `traces_v0.agent_input`, so the live value renders identically to the fresh-fetch path.

## Empty-value guard

A span-batch `trace_update` (before extraction runs) carries no `agent_input`. The frontend only overwrites `agentInput` when the payload value is non-empty, so an earlier batch can't blank a value a later extraction flush already set. The traces-table path keeps `rootSpanInput` as a transient fallback (unchanged behavior when the real value isn't present yet).

## Touch-points

**Backend**
- `app-server/src/traces/realtime.rs` — `agent_input: Option<String>` on both realtime structs; `from_trace` takes the value.
- `app-server/src/traces/processor.rs` — build `agent_input_by_trace` from `raw_trace_io`, thread it into `dispatch_trace_realtime_updates`.

**Frontend**
- `lib/traces/types.ts` — `agentInput` on `RealtimeTracePayload`.
- debugger store + content dispatcher — set `row.agentInput` in the known-run branch with the empty guard.
- `traces-table/realtime.ts` — prefer the real `agentInput`, fall back to `rootSpanInput`.

## Testing
- `cargo check` (default features) ✅
- `pnpm type-check` ✅ / `pnpm lint` (0 errors) ✅
- Manual: start a fresh debugger run, confirm the input pill appears live a few seconds after the trace (output appears immediately) with no reload.

---

> [!NOTE]
> **Low Risk**
> Additive SSE field and guarded client merges; no auth or persistence semantics change beyond richer realtime payloads.
> 
> **Overview**
> **Extracted user-task input (`agent_input`) now rides on live `trace_update` SSE payloads**, so debugger sessions and the traces table can show the input pill without a reload.
> 
> On ingest, the server builds `agent_input_by_trace` from `raw_trace_io` (JSON-stringified to match stored `agent_input`) and passes it into `dispatch_trace_realtime_updates`. **`RealtimeTrace` and `RealtimeDebuggerTrace` gain an `agentInput` field** populated only on the flush where input extraction landed; earlier span-batch updates omit it.
> 
> The frontend treats **`agentInput` as sticky**: merges prefer non-empty extracted values and ignore empty payloads so a late extraction update cannot be wiped by an earlier batch. The traces table still uses **`rootSpanInput` as a transient fallback** until the real value arrives; row updates layer extracted → prior → stand-in.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f82e3018728ba46e23bb60405da1d89ba1b0aa91. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

<a href="https://cubic.dev/pr/lmnr-ai/lmnr/pull/2153?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/app-server/src/traces/processor.rs
+++ b/app-server/src/traces/processor.rs
@@ -384,6 +384,21 @@ pub async fn process_span_messages(
         });
     }
 
+    // Extracted input threaded to the realtime dispatch as a first-class field
+    // (Option B in the handoff) so the debugger / traces table show it live —
+    // NOT sourced from the deprecated `lmnr_user_task` metadata fold above, which
+    // is on a demolition schedule. `value.to_string()` matches
+    // `CHTraceAgentInput.value` (and thus `traces_v0.agent_input`), so the live
+    // value renders identically to the fresh-fetch path.
+    let agent_input_by_trace: HashMap<(Uuid, Uuid), String> = raw_trace_io
+        .iter()
+        .filter_map(|io| {
+            io.input
+                .as_ref()
+                .map(|v| ((io.project_id, io.trace_id), v.to_string()))
+        })
+        .collect();
+
     // Enrich spans with usage info
     let mut span_usage_vec = Vec::with_capacity(messages.len());
 
@@ -736,7 +751,13 @@ pub async fn process_span_messages(
 
             debugger_session_blocks::upsert_blocks_for_traces(&db.pool, &updated_traces).await;
 
-            dispatch_trace_realtime_updates(&updated_traces, cache.clone(), &pubsub).await;
+            dispatch_trace_realtime_updates(
+                &updated_traces,
+                &agent_input_by_trace,
+                cache.clone(),
+                &pubsub,
+            )
+            .await;
         }
 
         // Dual-write partial rows to `traces_agg` (AggregatingMergeTree,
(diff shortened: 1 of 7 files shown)


olzhik11 (approved):
👍🏿

PR #2154, opened 2026-07-29
fix(realtime): remove wildcard CORS headers from SSE proxy route

okxint:
## Problem

`app/api/projects/[projectId]/realtime/route.ts` returns two response headers that shouldn't be there:

```
Access-Control-Allow-Origin: *
Access-Control-Allow-Headers: Cache-Control
```

## Why they're wrong

The `EventSource` that connects to this route is created in `lib/hooks/use-realtime.ts` at the same-origin path `/api/projects/${projectId}/realtime`. Same-origin requests don't need CORS headers — browsers only enforce CORS on cross-origin requests.

`Access-Control-Allow-Origin: *` is also potentially harmful: it would allow any cross-origin page to read the SSE stream for any project ID, bypassing the browser's same-origin protection for session-credentialed connections.

## Fix

Remove both headers. Every other header in the response (`Content-Type`, `Cache-Control`, `Connection`, `X-Accel-Buffering`) is correct and necessary for SSE to work.

<a href="https://cubic.dev/pr/lmnr-ai/lmnr/pull/2154?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

---

> [!NOTE]
> **Low Risk**
> Same-origin SSE no longer advertises permissive CORS; labeling targets change only for OpenAI message-array outputs and align with the trace UI.
> 
> **Overview**
> Removes **`Access-Control-Allow-Origin: *`** and **`Access-Control-Allow-Headers: Cache-Control`** from the project realtime SSE proxy response. The client connects same-origin via `EventSource`, so those headers were unnecessary and could widen cross-origin access to the stream.
> 
> When pushing a span to a labeling queue, **`pushSpanToLabelingQueue`** now sets the queue item **`target`** via **`normalizeSpanOutput`**: for OpenAI-style message arrays it takes the last assistant message’s **`content`**, JSON-parses string content when possible, and otherwise leaves output unchanged—so labelers see the same assistant payload as the trace view instead of the raw message array.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7dc467de5905bbc16d860a0242239414b62ce1cf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/frontend/app/api/projects/[projectId]/realtime/route.ts
+++ b/frontend/app/api/projects/[projectId]/realtime/route.ts
@@ -89,8 +89,6 @@ export async function GET(
         "Cache-Control": "no-cache",
         "Connection": "keep-alive",
         "X-Accel-Buffering": "no",
-        "Access-Control-Allow-Origin": "*",
-        "Access-Control-Allow-Headers": "Cache-Control",
       },
     });
   } catch (error) {
--- a/frontend/lib/actions/span/index.ts
+++ b/frontend/lib/actions/span/index.ts
@@ -150,6 +150,24 @@ export async function exportSpanToDataset(input: z.infer<typeof ExportSpanSchema
   });
 }
 
+// Unwrap OpenAI-style message arrays so the labeling queue target matches the
+// content shown in the trace view (which runs output through transformMessages).
+function normalizeSpanOutput(output: unknown): unknown {
+  if (!Array.isArray(output) || output.length === 0) return output;
+  const isOpenAIMessages = output.every(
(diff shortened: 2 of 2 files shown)


okxint:
Closing — targeting higher-traffic repos.

PR #2188, opened 2026-08-06
feat/blog styling

kolbeyang:
## Blog styling

Restyles blog and article posts. The main pieces:

- `app/typeset.css` — a new typeset stylesheet that owns article prose (headings, lists, tables, spacing). The MDX component map keeps only what needs bespoke behaviour.
- `blog-image.tsx` and `shared-trace-chip.tsx` — new components. Shared-trace links in prose render as an inline chip; `lib/blog/trace-links.ts` resolves which trace ids are actually public, so an unshared trace falls back to a plain link instead of a chip that 404s.
- `md-heading.tsx`, `post-layout.tsx`, `post-content/index.tsx` — heading anchors, two-column layout, and MDX overrides updated to match.

## Analytics: in-article link clicks

`post-layout.tsx` adds one attribute to the article element:

```
data-ph-capture-attribute-link_surface="blog_article"
```

PostHog promotes `data-ph-capture-attribute-*` from any ancestor of a clicked element onto the `$autocapture` event, so this labels every in-article link click without a listener. It also covers the raw HTML Strapi emits, which never routes through the MDX components map.

Clicks were already being captured with their href and text. This only makes in-article ones separable from header, footer, sidebar, and TOC links.

## Orange favicon in development

A local tab and a production tab look the same today. This makes them easy to tell apart at a glance. The dev icon is the same circular mark with a `primary-400` background (`#D0754E`), generated from `favicon.ico` itself so the shape and anti-aliasing match.

```ts
// app/layout.tsx
icons: {
  icon: process.env.NODE_ENV === "development" ? "/favicon-dev.png" : "/favicon.ico",
},
```

**Why both icons move to `public/`:** Next.js always prepends an `app/favicon.ico` to `metadata.icons`, even when you set that field yourself. An icon left in `app/` therefore wins over the conditional one, and you get two `<link rel="icon">` tags whose precedence is browser-dependent. This is a known upstream issue — [vercel/next.js#55767, favicon file overwrites icons defined in metadata](https://github.com/vercel/next.js/issues/55767) — and the accepted fix is to remove the file from `app/`. `public/favicon.ico` still serves at `/favicon.ico`, so crawlers hitting the well-known path are unaffected.

Keys on `NODE_ENV === "development"`, so `next dev` only. Staging runs `next start` and keeps the production icon.

Verified against the local dev server — one icon tag, no competing `favicon.ico`:

```
$ curl -s localhost:3010/sign-in | grep -o '<link rel="icon"[^>]*>'
<link rel="icon" href="/favicon-dev.png"/>
```

Favicons cache hard, so a hard reload may be needed to see it locally.


kolbeyang:
bugbot run

kolbeyang:
bugbot run

PR #2189, opened 2026-08-06
feat(frontend): use an orange favicon in development

kolbeyang:
A local tab and a production tab look the same today. This makes them easy to tell apart at a glance.

The dev icon is the same circular mark with a `primary-400` background (`#D0754E`), generated from `favicon.ico` so the shape and anti-aliasing match exactly.

| Production | Development |
|---|---|
| dark circle, white D | orange circle, white D |

## Why both icons move to `public/`

Next.js always prepends an `app/favicon.ico` to `metadata.icons`, even when you set that field yourself. An icon left in `app/` therefore wins over the conditional one, and you get two `<link rel="icon">` tags whose precedence is browser-dependent.

This is a known upstream issue: [vercel/next.js#55767 — favicon file overwrites icons defined in metadata](https://github.com/vercel/next.js/issues/55767). The accepted fix is to remove the file from `app/`, which is what this PR does. `public/favicon.ico` still serves at `/favicon.ico`, so crawlers that request the well-known path are unaffected.

## Scope

The switch keys on `NODE_ENV === "development"`, so it applies to `next dev` only. Staging runs `next start` and keeps the production icon. Say the word if staging should be orange too — it is a one-line change to the `ENVIRONMENT` variable.

## Verification

Ran against the local dev server:

```
$ curl -s localhost:3010/sign-in | grep -o '<link rel="icon"[^>]*>'
<link rel="icon" href="/favicon-dev.png"/>

$ curl -sI localhost:3010/favicon-dev.png
HTTP 200  image/png  3009B
```

One icon tag, no competing `favicon.ico`. Fill sampled at exactly `rgb(208,117,78)` = `primary-400`.

Note favicons cache hard — a hard reload may be needed to see the change locally.

<a href="https://cubic.dev/pr/lmnr-ai/lmnr/pull/2189?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


kolbeyang:
Folding this into #2188 instead — same review.

PR #2246, opened 2026-08-25
chore(frontend): replace ESLint + Prettier with oxlint + oxfmt

kolbeyang:
Replaces the frontend's ESLint + Prettier toolchain with the oxc equivalents. Two tools instead of three; **85 packages removed**.

| | before | after |
|---|---|---|
| lint | `eslint` + 7 plugins | `oxlint` |
| format | `prettier` | `oxfmt` |
| import sort | `eslint-plugin-simple-import-sort` | `oxfmt` `sortImports` |
| `pnpm lint` | 14.4s | **1.5s** |
| `pnpm format:check` | 4.6s | **~90ms** |

## Notes for review

- **41 source files reformatted.** Almost all are import reordering; the rest are union-type alias wraps. No logic changes. Consider adding the format commit to `.git-blame-ignore-revs`.
- **oxfmt replaces Prettier entirely**, not just for JS/TS — it also handles JSON, MD, MDX, CSS, SCSS, YAML, HTML, and GraphQL. Verified it produces byte-identical output to Prettier on every file Prettier previously owned (`components.json`, `tsconfig.json`, `README.md`, `initial-data.json`).
- **Existing `eslint-disable` comments still work.** oxlint accepts the legacy `react-hooks/exhaustive-deps` plugin name as an alias, so all 26 in-tree suppressions are untouched.
- **Ignore rules carried over.** `lib/policies/content/` and `assets/blog/` moved from `.prettierignore` into `oxfmtrc.ignorePatterns` — oxfmt reproduces Prettier's MDX emphasis rewrite (`*em*` to `_em_`) and trailing-comma injection into fenced code blocks, so the same protection is still required. `lib/db/migrations/` is excluded so drizzle-kit output stays canonical. `pnpm-lock.yaml` is excluded via `.gitignore`.
- **Versions pinned exact** (`oxlint@1.80.0`, `oxfmt@0.65.0`). oxfmt is pre-1.0, so formatting could shift across minor releases — intentional, don't widen to `^`.

## Verification

- `pnpm build` — passes, no warnings
- `npx tsc --noEmit --incremental false` — clean
- `pnpm check:circular` — clean
- `pnpm format:check` / `pnpm lint` — clean
- husky pre-commit hook exercised end-to-end
- `pnpm test` — 269/271 pass. **The 2 failures are pre-existing**, confirmed by running the suite at the merge-base (`8e99035`) in a scratch worktree: identical failures in `parseAiSdkMessages` and `normalizeToMessages`, unrelated to this change.

[...]

PR #2247, opened 2026-08-25
perf(frontend): require trace_id alongside span_id filters on spans

kolbeyang:
> **Stacked on #2246** (`feat/oxlint`). Review that one first; this PR's diff is only the last commit.

`spans` is `ORDER BY (project_id, trace_id, start_time, span_id)`. A `span_id` filter with no `trace_id` cannot prune the primary key — ClickHouse falls back to a generic exclusion search over every granule in the project and decompresses the heavy `input`/`output`/`attributes`/`events` columns for all of them.

`docs/internal/clickhouse-traces.md` already documents this, with numbers: **67/67 granules and ~1.3s vs 27/67 and ~0.4s** on a tiny dataset, and the gap widens with project size. It calls this "the dominant cost on the trace/span read routes." The guidance existed; nothing enforced it.

This adds an oxlint JS plugin that flags the pattern, plus the fixes.

## The fixes

Three call sites had the trace ids in hand and threw them away:

| Site | What was wrong |
|---|---|
| `getSpanTypes` | `attachSpanTypes` calls `parseSpanLinks`, which returns `{ label, traceId, spanId }` — the `traceId` was parsed and dropped by `.map(link => link.spanId)`. Now collected and passed. |
| `buildSpansQueryWithParams` / `…CountQueryWithParams` | New optional `traceIds` pairs into the `span_id` predicate. **Behaviour is unchanged when omitted**, so callers that scope traces structurally are unaffected. |
| `getSpans` | `spanHits` is typed `{ trace_id: string; span_id: string }[]`; the trace ids were mapped away. Now deduped and passed. |

Link filtering in `attachSpanTypes` tightened to require *both* ids to be well-formed UUIDs, since a link with only a span id can no longer be scoped.

## The suppressions

Four sites keep the old shape, each with an inline reason:

- **`addTagToCHSpan` / `removeTagFromCHSpan`** — `ALTER TABLE … UPDATE` mutations, not read-path granule scans.
- **`getSpanTags`** — reached from `GET /projects/:projectId/spans/:spanId/tags`, whose URL contract carries no trace id. `span.traceId` *is* available at the `<SpanTagsList>` call site, so this is fixable, but it changes an API route contract — deliberately out of scope here.
- **`getSpan`** — already appends `trace_id` when the caller supplies one; two internal callers (`exportSpanToDataset`, `pushSpanToLabelingQueue`) arrive from span-scoped routes without one.

I verified the suppressions are load-bearing: deleting one makes the rule fire again.

## Verification

- `oxlint .` — **0 errors** (rule active; confirmed it still fires on a synthetic violation)
- `tsc --noEmit --incremental false` — clean
- `oxfmt --check .` — clean
- `check:circular` — clean
- husky pre-commit hook ran end-to-end
- `pnpm test` — 269/271. The 2 failures (`parseAiSdkMessages`, `normalizeToMessages`) are **pre-existing**, confirmed at merge-base `8e99035`.

## Notes

- Adds `@oxlint/plugins@1.80.0` (exact pin, matching `oxlint`). JS plugins are the non-native path, but this one is a string-literal shape match — full-repo lint is unchanged at ~1.5s.
- The rule reads SQL string/template literals, so it cannot see trace scoping applied structurally (a `filters` entry or `customConditions`). That is why the shared builder carries a suppression rather than the rule attempting dataflow.
- Not verified against a real ClickHouse instance — the granule numbers quoted are from the existing internal doc, not re-measured here.

<a href="https://cubic.dev/pr/lmnr-ai/lmnr/pull/2247?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

---

> [!NOTE]
> **Medium Risk**
> Changes ClickHouse read queries and tightens event span-link handling (both UUIDs required), which could alter batch type lookups or empty results if callers omit trace ids; lint rule may need ongoing suppressions for edge cases.
> 
> **Overview**
> Adds an **oxlint** custom rule (`lmnr-clickhouse/span-id-needs-trace-id`) that errors on SQL string/template literals that filter `spans` by `span_id` without any `trace_id`, because that pattern prevents ClickHouse primary-key pruning on `ORDER BY (project_id, trace_id, …)`.
> 
> **Query fixes** where trace ids were already available but not used: `getSpanTypes` now requires `traceIds` and uses `trace_id IN … AND span_id IN …`; `attachSpanTypes` collects trace ids from span links (links must have both UUIDs); `getSpans` passes deduped trace ids from search hits into `buildSpansQueryWithParams` / count builders via new optional `traceIds` and shared `spanIdConditions`.
> 
> **Documented suppressions** for mutations (`ALTER TABLE … UPDATE`), span-only URL routes (`getSpan`, `getSpanTags`), and builders that scope traces via `filters` / `customConditions` instead of inline SQL. Adds `@oxlint/plugins` and registers the plugin in `.oxlintrc.json`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 85ab93e3a717122e687d6298b04be58ab7ea951e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/frontend/.oxlint-plugins/span-id-needs-trace-id.ts
+++ b/frontend/.oxlint-plugins/span-id-needs-trace-id.ts
@@ -1,0 +1,74 @@
+// Require a `trace_id` predicate alongside every `span_id` filter on `spans`.
+//
+// `spans` is ORDER BY (project_id, trace_id, start_time, span_id). A `span_id`
+// filter without `trace_id` cannot prune on the primary key: ClickHouse falls
+// back to a generic exclusion search over every granule in the project and
+// decompresses the heavy input/output/attributes/events columns for all of
+// them. Measured locally this was 67/67 granules and ~1.3s vs 27/67 and ~0.4s
+// on a tiny dataset, and the gap widens with project size. See
+// docs/internal/clickhouse-traces.md.
+//
+// Flagged: a string or template literal that filters on `span_id` (`=` or `IN`)
+// against `spans` and mentions no `trace_id` anywhere in the same literal.
+//
+// Not flagged:
+//   - literals that already mention `trace_id`
+//   - `DELETE FROM spans` — a mutation, not a granule scan on the read path
+//   - a query whose trace scoping arrives as a structured filter rather than
+//     inline SQL; those cannot be seen from the literal, so call sites that
+//     scope via `filters`/`customConditions` carry an eslint-disable with a
+//     one-line reason.
+import { eslintCompatPlugin } from "@oxlint/plugins";
+
+const SPAN_ID_PREDICATE = /\bspan_id\s*(?:=|\bIN\b|\bin\b)/i;
+const TRACE_ID = /\btrace_id\b/i;
+const FROM_SPANS = /\b(?:FROM|TABLE)\s+spans\b/i;
+const DELETE_FROM_SPANS = /\bDELETE\s+FROM\s+spans\b/i;
+
+// The raw text of a string or template literal, with `${...}` holes joined by a
+// space so a predicate split across an interpolation still reads as one clause.
+const literalText = (node: any): string | null => {
+  if (node?.type === "TemplateLiteral") {
+    return (node.quasis ?? []).map((q: any) => q?.value?.raw ?? "").join(" ");
+  }
+  if (node?.type === "Literal" && typeof node.value === "string") {
+    return node.value;
+  }
+  return null;
+};
+
+export default eslintCompatPlugin({
+  meta: { name: "lmnr-clickhouse" },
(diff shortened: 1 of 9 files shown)


kolbeyang:
Maybe later...

PR #2286, opened 2026-09-05
fix: stream large evaluation datapoint updates outside SQL

gregpr07:
## Problem

An evaluation can finish its executor and judge, then fail when publishing the final datapoint to Laminar:

```text
Clickhouse evaluation datapoint update failed:
Code: 62. DB::Exception: Max query size exceeded
```

This is a **result-publication failure**, not a model or browser failure. In the observed runs, execution had already incurred its cost, but the final output and scores could not replace the pre-created datapoint. Retrying the same update does not help: it generates the same oversized SQL statement.

### Why the existing limit does not protect this path

The update uses `INSERT ... SELECT` to preserve the existing row while changing output, trace, and scores. However, the Rust ClickHouse client's `.bind()` expands values into escaped SQL literals. For example:

```sql
if(empty(?), existing.executor_output, ?)
```

The output is bound **twice** in that expression; scores are bound **three times** elsewhere in the same query. The existing truncation caps output at approximately 250,000 bytes, but ClickHouse's default 262,144-byte limit applies to the **whole SQL query**, not each bound value.

A synthetic 200,000-byte ASCII output therefore becomes roughly 400,000 bytes of SQL literals, plus the rest of the statement, and fails even though it is below the truncation threshold. Escaping quotes/backslashes and embedding scores can grow the query further. Truncation also risks cutting serialized JSON mid-value, so merely shrinking the cap would lose evaluation evidence.

## Change

Keep the server-side `INSERT ... SELECT` merge, but stream one `JSONEachRow` record through `input()` using the existing ClickHouse client. Output, scores, and group values travel in the request body as data; only the three UUID filters remain SQL binds. This keeps SQL size independent of output size without raising the global query limit or fetching the existing row into application memory.

Remove the now-unnecessary output truncation. The existing HTTP request-size limit still applies; this is not an unlimited-payload change.

`API normalization -> JSONEachRow request body -> input() x scoped FINAL row -> INSERT`

## Proof

Three added tests passed using the **actual changed Rust writer module**, compiled in an isolated test crate with clickhouse 0.15.1, against a disposable local ClickHouse **26.5.7.64**:

- Old repeated-literal pattern: reproduced `Max query size exceeded` with a 200,000-byte output. New writer: same output round-trips.
- 2,800,013-byte JSON output with quotes, backslashes, newlines, and Unicode, plus a >300 KB scores map: exact output survives; scores merge correctly.
- Empty output/scores and nil trace preserve existing values; identical retry preserves the logical row; dataset links and other immutable fields survive; same datapoint UUID in other projects/evaluations stays unchanged; missing row is not inserted.
- HTTP-level tests verify payloads stay out of SQL/URL, SQL stays <4 KiB, URL stays <8 KiB, no `max_query_size` override, and write failures propagate.
- Per-file rustfmt and `git diff --check` pass.

Full backend `cargo check --locked` is **blocked locally**: installed rustc is 1.90.0; locked AWS dependencies require 1.94.1 and sqlx requires 1.94.0. Full backend tests/API/UI end-to-end tests were **not run**. The isolated tests are not a substitute for the backend CI gate.

To run the database regression with a compatible Rust toolchain:

```sh
docker run --rm -p 127.0.0.1:18123:8123 -e CLICKHOUSE_SKIP_USER_SETUP=1 clickhouse/clickhouse-server:26.5
# In another terminal, from app-server:
cargo test evaluations::datapoint_update::tests -- --include-ignored --nocapture
```

The database test uses only loopback and a UUID-named disposable database. Fixtures are synthetic; no production data or credentials are included.

## Compatibility / risk

- No schema, API, configuration, global query-limit, or dependency changes. Existing missing-datapoint checks, shared-trace handling, null-score filtering, and partial-update semantics remain in place.
- Full accepted outputs are now stored instead of silently clipped. This increases storage and per-request memory relative to truncated output; the existing HTTP payload limit (5 MiB by default) still applies. Realtime output previews remain capped at 200 characters.
- The merge still reads the existing row inside ClickHouse. This does not add a client-side read/merge/write or claim to solve pre-existing concurrent updates to the same datapoint.
- Default-compression database path tested. Managed/staging ClickHouse and the complete API/Postgres/shared-evaluation path still need maintainer validation.

## Rollout / rollback / recovery

Draft for review and CI. Maintainers should validate/backport to the deployment source (`lmnr-private`) before deploying; this PR does not change the hosted service. No migration is required. Rollback is an application revert; existing stored values remain compatible, but old large-update failures/truncation return.

After deployment, failed datapoint publications can be retried from retained results without rerunning the model. This PR does not replay or alter existing evaluation data.

<a href="https://cubic.dev/pr/lmnr-ai/lmnr/pull/2286?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

---

> [!NOTE]
> **Medium Risk**
> Changes how evaluation results are persisted in ClickHouse and removes silent output truncation, affecting storage size and a critical publish path; merge semantics are intended to stay the same but need CI/staging validation.
> 
> **Overview**
> Fixes evaluation datapoint publication failing with **Max query size exceeded** when executor output or scores are large. The old path inlined output and scores multiple times through `.bind()`, so payloads well under the prior ~250KB truncation cap could still blow past ClickHouse’s ~256KB SQL limit.
> 
> **Evaluation datapoint updates** now go through a dedicated `datapoint_update::write` that keeps the same `INSERT … SELECT` merge semantics (preserve immutable columns, empty incoming values keep existing fields, `jsonMergePatch` for scores) but feeds variable-size fields through ClickHouse `input()` and a **JSONEachRow** request body via `insert_formatted_with`. Only the datapoint/project/evaluation UUID filters stay as SQL binds.
> 
> Removes **250KB output truncation** (`value_to_ch_string_trunc` → full `value_to_ch_string`), so full executor output can be stored within existing HTTP limits. Adds wiremock tests for payload placement and error propagation, plus an optional local ClickHouse round-trip test for large payloads and merge/partial-update behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 89564b6388a46e005c3d898dffa66f269a5ebfdb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/app-server/src/evaluations/datapoint_update.rs
+++ b/app-server/src/evaluations/datapoint_update.rs
@@ -1,0 +1,79 @@
+use anyhow::Result;
+use clickhouse::Client;
+use serde::Serialize;
+use uuid::Uuid;
+
+#[derive(Serialize)]
+pub(super) struct Update<'a> {
+    pub trace_id: Uuid,
+    pub updated_at: i64,
+    pub executor_output: &'a str,
+    pub group_id: &'a str,
+    pub scores: &'a str,
+}
+
+pub(super) async fn write(
+    clickhouse: &Client,
+    evaluation_id: Uuid,
+    project_id: Uuid,
+    datapoint_id: Uuid,
+    update: &Update<'_>,
+) -> Result<()> {
+    // Stream variable-size values as data, not repeated SQL literals subject to max_query_size.
+    let query = clickhouse
+        .query(
+            "INSERT INTO evaluation_datapoints (
+                id, evaluation_id, project_id, trace_id, updated_at,
+                data, target, metadata, executor_output, `index`,
+                dataset_id, dataset_datapoint_id, dataset_datapoint_created_at,
+                group_id, scores
+            )
+            SELECT
+                existing.id,
+                existing.evaluation_id,
+                existing.project_id,
+                if(empty(incoming.trace_id), existing.trace_id, incoming.trace_id),
+                fromUnixTimestamp64Nano(incoming.updated_at, 'UTC'),
+                existing.data,
+                existing.target,
+                existing.metadata,
+                if(empty(incoming.executor_output), existing.executor_output, incoming.executor_output),
+                existing.`index`,
+                existing.dataset_id,
+                existing.dataset_datapoint_id,
+                existing.dataset_datapoint_created_at,
+                incoming.group_id,
+                if(empty(incoming.scores),
+                    existing.scores,
+                    if(notEmpty(existing.scores),
+                        jsonMergePatch(existing.scores, incoming.scores),
+                        incoming.scores))
+            FROM input('trace_id UUID, updated_at Int64, executor_output String, group_id String, scores String') AS incoming
+            CROSS JOIN (
+                SELECT
+                    id, evaluation_id, project_id, trace_id,
+                    data, target, metadata, executor_output, `index`, scores,
+                    dataset_id, dataset_datapoint_id, dataset_datapoint_created_at
+                FROM evaluation_datapoints FINAL
+                PREWHERE id = ?
+                WHERE project_id = ? AND evaluation_id = ?
+            ) AS existing
+            FORMAT JSONEachRow",
+        )
+        .bind(datapoint_id)
+        .bind(project_id)
+        .bind(evaluation_id);
+
+    let mut insert = clickhouse.insert_formatted_with(query.sql_display().to_string());
+    let payload = serde_json::to_vec(update)?;
+    async {
+        insert.send(payload.into()).await?;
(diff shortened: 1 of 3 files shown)


dinmukhamedm:
continued in #2287

PR #2289, opened 2026-09-06
fix(clusters): simplify chart axes and tooltip alignment

kolbeyang:
## Summary
- hide the right-side Signal runs axis while preserving its independent overlay scale
- align the Signal runs tooltip icon with cluster icon containers

## Verification
- seeded local project `clusters` with 54 traces, 54 signal runs, 54 events, and 5 hierarchical clusters
- verified with `agent-browser` that the chart renders one visible Y-axis
- verified the hovered tooltip contains Signal runs and cluster rows with aligned icon footprints
- `pnpm type-check`
- pre-commit: oxfmt, oxlint, circular dependency check

Screenshots were captured during browser verification and shared separately rather than committed to the repository.

---

> [!NOTE]
> **Low Risk**
> Presentation-only chart and tooltip styling with no auth, data, or API changes.
> 
> **Overview**
> **Time series chart:** When an overlay series is present, the right-hand overlay Y-axis is now **hidden** (`hide`) instead of showing a second tick scale. The overlay axis is kept so the area series still uses its own scale and isn’t compressed by the bar axis; values remain in the tooltip.
> 
> **Cluster stacked chart:** The **Signal runs** legend/tooltip icon is wrapped in a `size-4` flex container so its footprint matches `ClusterIcon` rows and labels line up in the tooltip.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c2be856a758e4f5a3b44eda3e264e75b773f74c2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


olzhik11 (approved):
👍🏿

PR #2317, opened 2026-09-09
data model (1)

olzhik11:
<a href="https://cubic.dev/pr/lmnr-ai/lmnr/pull/2317?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

---

> [!NOTE]
> **High Risk**
> Changes the security-boundary SQL rewriter and a mandatory ClickHouse view contract; incorrect eval_ids widening/narrowing would silently drop or over-scan evaluation rows, and migration 63 must ship with the app-server rewrite.
> 
> **Overview**
> Adds **predicate pushdown** for `evaluation_datapoints` queries by parameterizing `evaluation_datapoints_v0` with `eval_ids`, `min_start_time`, and `max_start_time` (migration 63), mirroring the existing `traces_v0` contract.
> 
> The SQL validator now rewrites `evaluation_datapoints` to always supply those three arguments: it derives **`eval_ids`** from safe `WHERE` shapes on `evaluation_id` (`=`, `IN`, AND/OR with the same conservative rules as time bounds), wraps scalars in `toUUID`, passes through `{name: Array(...)}` binds without double-wrapping, and uses **`[]`** when narrowing would be incorrect (`!=`, `NOT IN`, mixed OR with non-eval filters, joins on another table’s `evaluation_id`, etc.). **`start_time` bounds** on that relation reuse the existing traces time-bound extraction (±3h padding).
> 
> Tests cover hoist/widen/pass-through behavior and block users from setting `eval_ids` directly. Internal query-engine docs describe the mandatory sentinels and widen-never-narrow rule.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 679773627db1f9c8ffefdb6580ab79e43d8e4451. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


app-server/src/query_engine/validator.rs
```diff
@@ -1463,6 +1480,134 @@ fn traces_time_bound_args(where_clause: Option<&Expr>, alias: &str) -> (Expr, Ex
...
+                    filter: None,
+                    null_treatment: None,
+                    over: None,
+                    within_group: vec![],
+                })
+            })
+            .collect(),
+        named: false,
+    })
+}
```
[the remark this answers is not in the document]
olzhik11: tested, not real issue

app-server/src/query_engine/validator/tests.rs
```diff
@@ -161,18 +161,178 @@ fn test_trace_outputs_time_column_is_start_time() {
...
+        "SELECT id FROM evaluation_datapoints \
+         WHERE evaluation_id != '0195b6e0-0000-7000-8000-000000000001'",
+        "SELECT id FROM evaluation_datapoints \
+         WHERE evaluation_id NOT IN ('0195b6e0-0000-7000-8000-000000000001')",
+        "SELECT id FROM evaluation_datapoints WHERE index > 5",
+        // A per-row column cannot become a scalar view argument.
+        "SELECT id FROM evaluation_datapoints WHERE evaluation_id = group_id",
+        // Another relation's evaluation_id must not be borrowed.
+        "SELECT e.id FROM evaluation_datapoints AS e \
+         WHERE other.evaluation_id = '0195b6e0-0000-7000-8000-000000000001'",
```
dinmukhamedm, line 309: super nit: can we also extend this test to include an identical clause but with `SELECT e.evaluation_id`?

app-server/src/query_engine/validator/tests.rs
```diff
@@ -161,18 +161,178 @@ fn test_trace_outputs_time_column_is_start_time() {
...
+    assert_eq!(
+        eval_ids_of(
+            "SELECT id FROM evaluation_datapoints \
+             WHERE evaluation_id = '0195b6e0-0000-7000-8000-000000000001' \
+             OR evaluation_id = '0195b6e0-0000-7000-8000-000000000002'"
+        ),
+        "[toUUID('0195b6e0-0000-7000-8000-000000000001'), \
+         toUUID('0195b6e0-0000-7000-8000-000000000002')]"
+    );
+}
```
dinmukhamedm, line 267: I think the `AND` case is not covered well by the test. I had to stop and think for a second as to why is the AND picking a shorter side, and whether that is correct. I am convinced now that it is, but can we add a t3st for it? Something of a shape:

```sql
SELECT ...
WHERE evaluation_id IN ('uuid-1', 'uuid-2', 'uuid-3')
AND evaluation_id IN ('uuid-3', uuid-4')
```

dinmukhamedm (approved):

PR #2320, opened 2026-09-09
fix: enum filters signal events

olzhik11:
> [!NOTE]
> **Medium Risk**
> Changes event query WHERE clauses for payload string/enum filters; incorrect logic would affect which events appear in the table.
> 
> **Overview**
> **Signal event payload enum fields** now register as `enum` filters with selectable options from the signal schema, instead of free-text string filters.
> 
> **Payload JSON string filtering** in ClickHouse query building fixes a bug where `!=` combined `simpleJSONExtractString` and `simpleJSONExtractRaw` with **OR**, which could match almost every row (quoted raw values never equal the unquoted filter value). Equality still uses **OR** so either extraction form can match; inequality now uses **AND** so both must disagree with the filter value.
> 
> A unit test asserts the generated SQL uses OR for `=` and AND for `!=` on payload string filters.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0f471bd9dfd70f55e24e62af634a7f7289a22314. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/frontend/components/signal/events-table/columns.tsx
+++ b/frontend/components/signal/events-table/columns.tsx
@@ -195,7 +195,8 @@ function createPayloadFilter(field: SchemaField): ColumnFilter {
       return {
         name: field.name,
         key: `payload.${field.name}`,
-        dataType: "string",
+        dataType: "enum",
+        options: (field.enumValues ?? []).map((v) => ({ label: v, value: v })),
       };
     default:
       return {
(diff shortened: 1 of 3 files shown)


kolbeyang (approved):
👍

PR #2338, opened 2026-09-11
fix(traces): prevent tool definition overflow

kolbeyang:
## Summary
- give the tool-definition dropdown a definite responsive width
- prevent Radix ScrollArea's intrinsic table wrapper from widening past the viewport
- wrap long tool names and constrain nested tool content

## Verification
- `npx tsc --noEmit --incremental false`
- `npx oxlint components/traces/tool-list.tsx`
- `npx oxfmt --check components/traces/tool-list.tsx`
- tested the seeded long-definition trace in `agent-browser` at 1280×800 and 800×700
- verified the dropdown viewport has `clientWidth === scrollWidth` and zero overflowing descendants

## Before

Tool definition popover before

## After

Tool definition popover after


olzhik11 (approved):
👍🏿

PR #2340, opened 2026-09-11
Add dataset CRUD API

kolbeyang:
## Dataset API

Available with project API-key auth under `/v1` and user-token auth under `/v1/cli`.

| Method | Path | Behavior |
|---|---|---|
| `POST` | `/datasets` | Create an empty dataset from `{ "name": string }`. |
| `GET` | `/datasets/{dataset_id}` | Get a dataset by UUID. |
| `PATCH` | `/datasets/{dataset_id}` | Rename a dataset from `{ "name": string }`. |
| `DELETE` | `/datasets/{dataset_id}` | Delete a dataset and its datapoints. |

Names must be non-empty, are stored unchanged, and do not need to be unique. Missing or cross-project dataset IDs return `404`.

`GET /datasets/datapoints` now requires `limit` from `1` to `1000` and a non-negative `offset`.


olzhik11 (approved):
👍🏿

PR #2346, opened 2026-09-13
fix(signals): polish empty clusters state

kolbeyang:
## Summary
- reduce the clusters chart height from 320px to 250px
- hide the clusters placeholder and regular breadcrumbs when the selected range has no clusters
- keep emerging-cluster breadcrumbs visible whenever `emergingClusterId` is active, including ranges without regular clusters
- restore `% of traces` in cluster tooltips and the selected-cluster readout
- show loading UI until cluster tree, cluster stats, and run stats settle, preventing mixed-window percentages
- keep the removed background run-count line graph out of the stacked chart

### Graph height and `% of traces`

| Before | After |
| --- | --- |
| Previous graph height and cluster details | Reduced graph height with trace percentage |

### No-clusters placeholder

| Before | After |
| --- | --- |
| No-clusters placeholder before | No-clusters placeholder hidden after |

### Emerging-cluster breadcrumbs without regular clusters

| Before | After |
| --- | --- |
| Emerging-cluster filter hidden before | Emerging-cluster filter visible after |

All screenshots use existing local database mock data, rendered through the running frontend and app server with `agent-browser`.

## Verification
- `npx tsc --noEmit --incremental false`
- frontend pre-commit checks: oxfmt, oxlint, circular dependency check, TypeScript
- browser-verified selected-cluster readout and hovered tooltip show `% of traces`
- browser-verified empty and emerging-cluster states

---

> [!NOTE]
> **Medium Risk**
> Refactors how signal cluster data is loaded and synced between SWR and the Zustand store; incorrect rangeKey or partial updates could briefly show mismatched chart vs. filter state, but scope is limited to the signals clusters UI.
> 
> **Overview**
> Introduces a **single cluster visualization snapshot** (`/clusters/visualization`) that loads clusters, time-series stats, unclustered counts, and **signal run totals** together, so the icicle strip, stacked chart, and event-table filter stay on the same time window instead of mixing separate fetches.
> 
> The clusters section now drives UI from that SWR snapshot (with `rangeKey` matching) and **pushes complete snapshots into the signal store** via `setClusters`, replacing store-owned `fetchClusters` and the old stats-only API. Refresh in the events table revalidates this endpoint only.
> 
> **Restores `% of traces`** in icicle tooltips and the selected-cluster readout, using `traceTotal` as the denominator (capped at 100% when one trace raises multiple events).
> 
> **Empty-state polish:** chart height **320px → 250px**; hides the icicle strip, regular breadcrumbs, and the old empty placeholder when the range has no clusters, while still showing **emerging-cluster breadcrumbs** when that filter is active; loading skeleton until the first snapshot for the current range is ready.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 69d23ca13920e104f763cbf53cf55d8db417f67d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/frontend/components/signal/clusters-section/cluster-icicle/band-details.tsx
+++ b/frontend/components/signal/clusters-section/cluster-icicle/band-details.tsx
@@ -26,13 +26,21 @@ const realChildCount = (node: DetailedNode) => (node.all ?? node.children).lengt
 /**
  * Everything a cluster says about itself besides its name.
  *
+ * `traceTotal` is the signal's run count over the window — the traces this
+ * signal actually evaluated — so the share reads against what was looked at
+ * rather than against what happened to be clustered.
+ *
  * No level: L1/L2/L3 is our word for how deep the tree is, not something anyone
  * reading the chart needs.
  */
-export function clusterFacts(node: DetailedNode): string[] {
+export function clusterFacts(node: DetailedNode, traceTotal: number): string[] {
+  // A single trace can raise several events, so the ratio is not bounded by one.
+  // Capped rather than shown, because ">100% of traces" reads as a broken number.
+  const share = traceTotal > 0 ? Math.min(100, (node.total / traceTotal) * 100) : 0;
   const childCount = realChildCount(node);
   return [
     `${node.total.toLocaleString()} events`,
+    share > 0 ? `${share < 1 ? "<1" : Math.round(share)}% of traces` : null,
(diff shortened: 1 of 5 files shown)


olzhik11 (approved):
👍🏿

PR #2357, opened 2026-09-14
pii(1/5): pii-redactor returns canonical text + masks; app-server splices

Rainhunter13:
## Summary

PR 1/5 of the PII masks + trace-level access policy stack (replaces #2334, split for review; the stack merges together 1 → 2 → 3 → 4 → 5). Design notes for the whole stack land in `docs/internal/rbac.md` (PR 4).

The `pii-redactor` stops returning redacted copies. It re-serializes each stringified-JSON text to a canonical compact form and returns it together with the byte ranges of the PII it found; the caller splices `[REDACTED_<LABEL>]` over those ranges. Behaviour of the existing `removePii` projects is unchanged: app-server splices at ingest and stores the redacted text, as before. Later PRs store the masks next to the raw text (`dual` mode) and splice in ClickHouse at read time.

## What changes

**pii-redactor (lockstep deploy with app-server)**
- `Redact` returns `results: [{ text, masks: [{start, end, label}] }]`. `text` is byte-equal to `serde_json::to_string` of the parsed input (nested stringified JSON compacted too); `masks` are byte ranges into exactly that string. `placeholder_format` is gone: the placeholder is fixed on the reader side. Old fields are `reserved`.
- `json_walker::serialize_with_masks` is a hand-written offset-tracking serializer: string leaves are escaped piecewise at span boundaries so mask offsets fall out of the output cursor; stringified inner JSON is serialized first and its masks translated through the outer escape.

**app-server**
- `RedactTexts` trait (the gRPC client is the production impl; tests plug in a fake), `MaskedText` / `PiiMask` with `apply_masks` (`pii_redactor/masks.rs`): masks must be sorted, non-overlapping, in bounds and on char boundaries; a malformed mask is an error, never partially redacted text.
- `redact_spans_in_place` sanitizes request texts before the RPC (what the redactor sees is what would have been stored), splices the masks and writes the result back to `span.input`/`output`, the `unique_content` rows and the Quickwit buffers. Any RPC / parse / mask failure leaves the affected buffer untouched, as before.
- Project opt-in resolution moves to `resolve_opted_in_projects` (called from the processor) so the state machine is unit-testable.

## Lockstep
The `pii-redactor` image must ship with the same rollout (response shape changed in place).

## Verification
- `cargo test` in `pii-redactor` (31, incl. byte-equality with `serde_json` on every fixture and nested/doubly-nested stringified JSON).
- `cargo test --bin app-server pii_redactor` (10), `cargo check` zero warnings.
- End-to-end against the real ONNX model was done on the full stack (#2334).

<a href="https://cubic.dev/pr/lmnr-ai/lmnr/pull/2357?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

---

> [!NOTE]
> **High Risk**
> Breaking gRPC contract and lockstep deploy on the PII ingestion path; incorrect mask handling could leak PII or drop redaction despite best-effort fallbacks.
> 
> **Overview**
> **pii-redactor** no longer returns pre-redacted JSON strings. Each `Redact` result is **`RedactedText`**: a compact canonical re-serialization plus sorted **byte-range masks** (and labels). **`placeholder_format` is removed**; callers splice fixed **`[REDACTED_<LABEL>]`** placeholders. Detection still walks stringified JSON the same way; **`serialize_with_masks`** replaces in-service rewriting with offset-tracking serialization (including nested stringified JSON and escaped bytes).
> 
> **app-server** maps gRPC results to **`MaskedText`**, validates masks strictly (malformed ranges fail the field, never partial redaction), and **`redact_spans_in_place`** splices before writing spans, **`unique_content`**, and Quickwit trace-new buffers. Span payloads are **sanitized before the RPC**; project opt-in moves to **`resolve_opted_in_projects`** in the trace processor for testability. A **`RedactTexts`** trait backs unit tests for the ingest path.
> 
> **Deploy:** **pii-redactor and app-server must roll out together** (response shape changed in place). Stored redacted output for `removePii` projects should match prior behavior; later stack PRs can persist masks instead of splicing at ingest.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a185b669c65c03cbdbc833f5981f3bdc680de081. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


app-server/src/pii_redactor/masks.rs
```diff
@@ -0,0 +1,154 @@
...
+                    start: m.start,
+                    end: m.end,
+                    label: m.label,
+                })
+                .collect(),
+        }
+    }
+}
+
+/// Masks must be sorted, non-overlapping, within `text`, and on char
```
dinmukhamedm, line 54: > non-overlapping

do we have a way to control this or does the model decide? Should we maybe throw away one/both sides of an overlap instead of failing the entire extraction?

UPD: I think we actually filter out in `normalize_ranges`, please confirm my understanding

app-server/src/pii_redactor/masks.rs
```diff
@@ -0,0 +1,154 @@
...
+/// boundaries; anything else is an error, never a best-effort splice.
+pub fn validate_masks(text: &str, masks: &[PiiMask]) -> Result<()> {
```
dinmukhamedm: nit: does this have to be pub?

From my old Java days: let's keep the public interfaces at the top of the file and move this function below them if it can be made private.

app-server/src/pii_redactor/masks.rs
Rainhunter13: yes, `normalize_ranges` handles overlaps, `validate_masks` is a safety on the other side

app-server/src/pii_redactor/masks.rs
Rainhunter13: Done in e237e50b8: `validate_masks` and `apply_masks` are both private now; nothing outside `masks.rs` calls them (PR 5 only adds `MaskedText` methods that delegate to them). The public surface is the two types and `impl MaskedText` at the top, helpers below.

dinmukhamedm (approved):

kolbeyang (approved):
👍

dinmukhamedm (approved):

PR #2358, opened 2026-09-14
pii(2/5): piiMode setting (off | redact | dual) replaces removePii

Rainhunter13:
## Summary

PR 2/5 of the PII masks + trace-level access policy stack (stacked on #2357; merges together 1 → 2 → 3 → 4 → 5).

`projects.settings.piiMode: off | redact | dual` replaces the `removePii` boolean. `off` and `redact` behave exactly as `false`/`true` did. `dual` (raw text stored with its PII masks, masked per role at read time) is accepted by the schema so the model is final, but the settings UI does not offer it yet and ingestion treats it like `redact` until PR 5 adds the storage.

## What changes

**Postgres**
- No migration. Both readers translate `removePii: true` to `redact` when `piiMode` is absent, and the key stays in the JSONB so an app-server pod still on the previous binary (which reads only `removePii`; the 7-day project cache carries the same shape) keeps redacting during the rollout. `updateProjectSettings` mirrors `piiMode` into `removePii` (`mode !== "off"`) so a change made through the new UI during that window reaches old pods too. A one-off migration can drop the key once the fallback is retired.

**app-server**
- `PiiMode` (`db/projects.rs`) with `ProjectSettings::pii_mode()` applying the `removePii` fallback. An unknown mode string (a frontend ahead of this binary) reads as `redact`, so a new mode can never silently disable redaction.
- `resolve_project_pii_modes` replaces the opt-in set; a failed settings lookup leaves the project's spans untouched (as before).

**frontend**
- `ProjectSettingsSchema.piiMode`, `parseStoredProjectSettings` (shared by the project loader and the settings route), Pro gate keyed on `mode !== "off"`. `PII_MODES` / `PiiMode` live in the client-safe `lib/actions/project/types.ts` (`settings.ts` pulls in the db and Redis clients, so a client component must not import it).
- `PATCH /api/projects/[projectId]/settings` requires workspace owner/admin to change `piiMode`; the `PiiRedaction` control is a select (off / on), disabled for members.

## Verification
- `cargo test --bin app-server pii_redactor` (12) and `db::projects` (3); `cargo check` zero warnings.
- `tsc --noEmit --incremental false` clean; oxlint/oxfmt clean on touched files.

---

> [!NOTE]
> **Medium Risk**
> Changes how PII redaction is configured and gated at ingestion (sensitive data handling), with rollout fallbacks but new auth rules and mode semantics that must stay consistent across frontend and app-server.
> 
> **Overview**
> **Replaces the `removePii` boolean with `projects.settings.piiMode` (`off` | `redact` | `dual`)** across app-server and frontend, with no DB migration.
> 
> **app-server:** Adds `PiiMode` and `ProjectSettings::pii_mode()` (legacy `removePii: true` → `redact`; unknown mode strings fail closed to `redact`). Ingestion redaction is driven by per-project mode via `resolve_project_pii_modes`; `redact` and `dual` both run the redactor today (`dual` until mask storage lands). Failed project settings lookups still skip redaction for those spans.
> 
> **frontend:** Settings schema and `parseStoredProjectSettings` use `piiMode`; Pro tier gates any mode other than `off`. Writes mirror `piiMode` into `removePii` for older app-server pods during rollout. The security UI is a mode select (off/on only in the picker); only workspace owners/admins may change `piiMode` (API + UI).
> 
> **Rollout:** Readers on both sides honor `removePii` when `piiMode` is absent; new writes keep both keys in sync.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5a2f08369598470895e40622d745c3f2a5bc845d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


frontend/lib/db/migrations/0111_pii_mode.sql
```diff
@@ -0,0 +1,13 @@
...
+-- ("off" | "redact" | "dual"). `removePii: true` becomes "redact"; an already
+-- present `piiMode` wins. Rows without the legacy key are untouched.
+UPDATE "projects"
+SET "settings" = ("settings" - 'removePii')
+  || CASE
+       WHEN NOT ("settings" ? 'piiMode') AND "settings"->>'removePii' = 'true'
+         THEN '{"piiMode":"redact"}'::jsonb
+       ELSE '{}'::jsonb
+     END
+WHERE "settings" ? 'removePii';
```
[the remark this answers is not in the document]
Rainhunter13: Fixed in c783380ce: the migration now only adds `piiMode: "redact"` where `removePii: true` and leaves `removePii` in place, so a pod on the previous binary (and the cached `project:{id}` shape it reads) keeps redacting through the rollout. Both readers ignore the key once `piiMode` is set; a later migration can drop it. Dry-run on local Postgres: `{removePii:true}` → `{piiMode:redact, removePii:true}`, existing `piiMode` wins, other rows untouched.

frontend/lib/actions/project/settings.ts
```diff
@@ -69,7 +90,10 @@ export async function updateProjectSettings(input: z.infer<typeof UpdateProjectS
   // gate doesn't apply there — skip it entirely off Laminar Cloud.
   const enablesGatedKey =
     isFeatureEnabled(Feature.LAMINAR_CLOUD) &&
-    (Object.keys(settings) as (keyof ProjectSettings)[]).some((k) => PRO_TIER_KEYS.has(k) && settings[k] === true);
+    (Object.keys(settings) as (keyof ProjectSettings)[]).some((k) => {
+      const gate = PRO_TIER_GATES[k] as ((value: unknown) => boolean) | undefined;
+      return gate !== undefined && settings[k] !== undefined && gate(settings[k]);
+    });
```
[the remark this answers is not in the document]
Rainhunter13: Fixed in 23a3d4144: `updateProjectSettings` now writes `removePii: piiMode !== "off"` alongside `piiMode`, so a pod still on the previous binary (which reads only `removePii`) follows changes made during the rollout in both directions. The project cache entry is already invalidated after the write, so old pods re-read the mirrored value on the next batch.

app-server/src/db/projects.rs
```diff
@@ -1,9 +1,36 @@
 use anyhow::Result;
 use chrono::{DateTime, Utc};
-use serde::{Deserialize, Serialize};
+use serde::{Deserialize, Deserializer, Serialize};
 use sqlx::{FromRow, PgPool, types::Json};
 use uuid::Uuid;
 
+/// How a project's span input/output is treated for PII.
+#[derive(Deserialize, Serialize, Default, Clone, Copy, Debug, PartialEq, Eq)]
+#[serde(rename_all = "lowercase")]
```
dinmukhamedm: super nit: should we maybe do "snake_case" just in case we have more than one word?

app-server/src/db/projects.rs
```diff
@@ -1,9 +1,36 @@
...
+    Ok(Option::<String>::deserialize(d)?.map(|s| match s.as_str() {
+        "off" => PiiMode::Off,
+        "redact" => PiiMode::Redact,
+        "dual" => PiiMode::Dual,
+        other => {
+            log::warn!("unknown piiMode {other:?}, treating as redact");
+            PiiMode::Redact
+        }
+    }))
+}
```
dinmukhamedm, line 32: will we explicitly set it to `off` for all non-pro projects manually?

Rainhunter13: it's backward compatible at the readers side – if `mode` is absent , we check the pre-existing `removePII` flag. If it's `on` then act as a `mode=redact`, otherwise as if `mode=off`

so nothing needed to set manually

app-server/src/db/projects.rs
Rainhunter13: Done in 1a9fa0841. No-op for the current single-word variants (`off`/`redact`/`dual` serialize identically), so nothing on the wire or in Postgres changes.

dinmukhamedm (dismissed):

dinmukhamedm (approved):

PR #2421, opened 2026-09-22
feat(llm-profiles): Responses API shape for custom gateway profiles

Rainhunter13:
Custom (OpenAI-compatible) LLM profiles only spoke Chat Completions. This adds a `custom_responses` provider so a custom gateway can be used over the Responses API (`/responses`). It follows the pattern the OpenAI and Azure families already use: the API shape is part of the provider name, and the form collapses the family into one entry with an API-shape select.

## Shape
- `custom` keeps meaning Chat Completions; `custom_responses` is new. Existing rows, cached profiles and API/CLI callers are unchanged, so no migration is needed.
- **app-server**: new `LlmProfileProvider::CustomResponses` (`reported_name` → `openai`). `is_custom_gateway()` covers both custom providers in validation and in the custom-header secret checks (`prune_secrets` / `assert_secrets_complete`), so header values survive a save for either shape. `build.rs` builds the `OpenAIResponsesClient` from the same base URL, key and headers for `custom_responses`.
- **frontend**: `CUSTOM_PROVIDERS` / `isCustomProvider` and one Zod branch for both custom providers. The form's "Custom (OpenAI-compatible)" entry gets an API-shape select (a `customShape` form value, mapped by `toProvider` / `toUiProvider`), and the stored key and headers are kept when only the shape changes. The base-URL hint follows the shape. The labels become "Custom (Chat Completions)" and "Custom (Responses)", in the settings table and the signal picker. `thinkingNamespace` maps `custom_responses` to `openai`.
- **`store: false` on the frontend Responses model.** `languageModelFromProfile` uses `.responses()` for `custom_responses`, wrapped with `defaultSettingsMiddleware` to set `store: false`. With the default `store: true`, the AI SDK sends earlier steps of a multi-step call back as `item_reference` ids, which a gateway that doesn't persist responses can't resolve. The middleware deep-merges, so call-level `providerOptions.openai` still apply. Only multi-step frontend features are affected (today `render-template/generate.ts`). The app-server never sends `store` or `item_reference`.

- **OSS: workspace profile routes are no longer behind the `signals` cargo feature.** `/api/v1/workspaces/{workspace_id}/llm-profiles` (create, update, delete, `test`) used to be registered only in `signals` builds. The OSS UI, which is always shown and used by the playground and LLM feature routes, got a bare 404 on every save and Test click. The CLI and project-API profile routes were already ungated, and the routes use the same `llm/profiles/service`, so no stubs were needed.

## Testing
- `cargo test llm::profiles`: 23 pass. New or extended tests cover the `custom_responses` round trip, `normalize_config` for `custom_responses`, header prune and completeness checks for both custom providers, client selection in `build.rs`, and a probe hitting `/responses` with the custom header.
- Frontend (`tsx --test`), 13 pass:
  - `llm-profile-model.test.ts` (stubbed `fetch`): `custom_responses` posts to `{baseUrl}/responses` with the custom header and `store: false`, call-level openai options merge on top, and `custom` posts to `/chat/completions`.
  - `llm-profile-form.test.ts`: the provider survives an edit round trip through the form for both shapes, and the stored secrets are kept when only the shape changes.
  - `llm-profile-schema.test.ts`: `custom_responses` validation.
- OSS build (default features): `POST .../llm-profiles/test` returns 200 with `ok: true` for `custom_responses` and `custom` through LiteLLM (`gpt-5.4-mini`) and for `openai_responses` direct (`gpt-5.6-luna`). Before this change it was a 404.
- `tsc --noEmit --incremental false`, oxlint, oxfmt, `cargo check` and the pre-commit hooks are clean.
- Live: a signal run through LiteLLM (`/v1/responses`, OpenAI and Gemini models behind it) via the env provider path this reuses, plus direct OpenAI with `gpt-5.6-luna`, all produced correct findings. The frontend model ran a two-step tool loop through LiteLLM to OpenAI successfully with `store: false`.

Known gateway limitation, not fixed here: LiteLLM's Gemini-to-Responses translation returns `"text": null` in `output_text`, which the AI SDK's response schema rejects on the frontend, with or without this change. The app-server's parser tolerates it.

---

> [!NOTE]
> **Medium Risk**
> Touches encrypted credential handling and LLM routing for custom gateways; the route-registration change affects all OSS workspace profile writes but is a deliberate exposure of existing handlers.
> 
> **Overview**
> Adds **`custom_responses`** as a new LLM profile provider so OpenAI-compatible gateways can be configured for the **Responses API** (`/responses`), while **`custom`** remains Chat Completions. Backend client build, validation, secret pruning, and connection probes branch on provider shape (Chat vs Responses) and share gateway config via **`is_custom_gateway()`**.
> 
> The workspace LLM profile UI mirrors OpenAI/Azure: one “Custom” entry with an API-shape select, round-trip for both wire names, and stored credentials kept when only the shape changes. Playground routing uses **`.responses()`** for `custom_responses` with **`store: false`** so multi-step AI SDK calls resend content instead of `item_reference` ids gateways cannot resolve.
> 
> **OSS fix:** internal workspace **`/api/v1/workspaces/{id}/llm-profiles`** routes are no longer gated behind the `signals` cargo feature, so default builds can create/update/probe profiles from the UI (previously 404).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit db03ca56cea177a12df94be15d9ed3caa96de100. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


olzhik11 (approved):
Image

dinmukhamedm (approved):

dinmukhamedm (approved):

PR #2424, opened 2026-09-23
fix(llm-profiles): reject fields a provider drops; surface a malformed AEAD_SECRET_KEY

Rainhunter13:
Two issues a self-hosting user reported while setting up LLM profiles through the CLI/API.

## 1. Unused fields were silently dropped
`azure_responses` with `--header` stored `headers: []` with no error. It went further than Azure: `normalize_config` rebuilt the config with only the fields each provider uses, and `prune_secrets` dropped header values for every non-gateway provider. So `openai_responses` also silently lost a `baseUrl`, and Gemini lost a `region`. The user thought they had configured something that was never sent.

Writes (create, update, and the Test probe) now return a 400 instead:
- `reject_unused_fields`: any `config` field the provider never sends (`region` off Bedrock, `resourceId`/`apiVersion` off Azure, `baseUrl` off Azure/custom, `headerNames` off custom). A blank string counts as absent.
- `validate_secrets` (the one check for incoming secrets, now also covering where headers go): header values for a non-gateway provider, or for a name not listed in `headerNames`.

The error names the unused fields (e.g. "Provider `azure_responses` does not use headerNames"). The reporter's actual need, a self-hosted gateway with Responses and an attribution header, is covered by `custom_responses`.

The UI is unaffected: `buildConfig` only sends the chosen provider's fields, and header values only for custom. Only hand-written API/CLI calls see the new 400s.

## 2. A malformed `AEAD_SECRET_KEY` gave an opaque 500
`crypto.rs` already produced a precise error, but the service wrapped it as `CrudError::Internal`, which returns `{"error":"Internal server error"}`.
- **Startup:** if `AEAD_SECRET_KEY` is set and malformed, the app-server logs the decode error at boot (`crypto::check_key`). Unset stays allowed, since only profiles, the data plane and Slack need it.
- **Profile service:** a new `CrudError::Misconfigured` returns that message in the 500 body. Writes and probes check the key before encrypting, and stored-secret decrypt failures use it too. The message never includes the key.
- **Frontend:** `lib/crypto.ts` requires exactly 64 hex characters (`Buffer.from(hex)` silently truncated a bad key, so libsodium failed later with an unhelpful error). The key is now loaded outside each function's `try`, so the reason isn't hidden behind "Failed to decode value…".

## Testing
- `cargo test llm::profiles`: 25 pass. New tests cover Azure with `headerNames` (asserts the message), `openai_responses` with `baseUrl`/`headerNames`, custom with `apiVersion`, blank-as-absent, and header values for a non-gateway provider or an unlisted name. The old "Gemini ignores `region`" test now expects a 400.
- `tests/crypto-key.test.ts` (new): a valid key round-trips; a short key, a non-hex character and a quoted key are all reported by name; an unset key is reported.
- Live against the OSS build:
  - with `AEAD_SECRET_KEY` wrapped in quotes (a common Docker env mistake): boot logs `Failed to decode AEAD_SECRET_KEY from hex: Invalid character '"' at position 0…`, and create/test return that message;
  - with a valid key: the exact reported request (`azure_responses` + `headerNames`) → 400 naming the field; header value only → 400; `openai_responses` + `baseUrl` → 400; probe → 400; `custom_responses` + header → stored with the header.
- `cargo check`, `tsc --noEmit --incremental false`, oxlint/oxfmt and the pre-commit hooks are clean.

Made with [Cursor](https://cursor.com)

<a href="https://cubic.dev/pr/lmnr-ai/lmnr/pull/2424?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


app-server/src/llm/profiles/service/provider_fields.rs
```diff
@@ -0,0 +1,83 @@
...
+
+    fn wire_name(self) -> &'static str {
+        match self {
+            Self::Region => "region",
+            Self::ResourceId => "resourceId",
+            Self::ApiVersion => "apiVersion",
+            Self::BaseUrl => "baseUrl",
+            Self::HeaderNames => "headerNames",
+        }
+    }
```
dinmukhamedm, line 35: nit: could just serialize with rename_all="camelCase" but whatever, I guess the serde serialization is an overkill

app-server/src/llm/profiles/service/validate.rs
```diff
@@ -156,6 +158,38 @@ pub(super) fn normalize_config(
...
+/// `headerNames`; anything else would be silently discarded on save.
+fn reject_unused_header_secrets(
+    provider: LlmProfileProvider,
+    config: &ProfileConfig,
+    secrets: &ProfileSecrets,
+) -> Result<(), CrudError> {
+    if secrets.headers.is_empty() {
+        return Ok(());
+    }
+    if !provider.is_custom_gateway() {
```
dinmukhamedm, line 171: should we not allow headers to regular providers? why?

I think of the headers as a per-request override

Rainhunter13: not sure if there is a use-case for that? ig the idea is to keep main providers simple and use custom provider for anything custom needed?

dinmukhamedm: Yeah, let's go with this, and we can allow headers when/if customers need it. I was just thinking about Anthropic that used to allow controlling experimental features via headers, but probably too narrow of a use case

dinmukhamedm (approved):

